PR #2, opened 2017-02-06
Fix O(n^2) response time for "list-security-groups" on openstack-nova.

geomacy:
Proposed fix for https://issues.apache.org/jira/browse/JCLOUDS-1235.
    
This change takes the approach of storing the information about the overall list of groups within the `SecurityGroupInRegion` when it is created, so that any subsequent conversion operation has access to all the groups in the same region as the one to be converted.
    
It also collapses the functionality of `NovaSecurityGroupToSecurityGroup`,
`SecurityGroupRuleToIpPermission` and `FindSecurityGroupWithNameAndReturnTrue`
all into `NovaSecurityGroupInRegionToSecurityGroup`. These three classes could arguably
be removed or at least deprecated; I haven’t done either in this PR yet and would ask you for guidance on the best approach here.

Another point for review, note the comment in `NovaSecurityGroupInRegionToSecurityGroup.java:110` about the workaround to the issue https://issues.apache.org/jira/browse/JCLOUDS-1234.  Is this the best approach in this circumstance?

### DIFF
--- a/apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/extensions/NovaSecurityGroupExtension.java
+++ b/apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/extensions/NovaSecurityGroupExtension.java
@@ -58,6 +58,7 @@ import com.google.common.base.Function;
 import com.google.common.base.Optional;
 import com.google.common.base.Supplier;
 import com.google.common.cache.LoadingCache;
+import com.google.common.collect.FluentIterable;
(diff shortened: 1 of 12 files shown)


geomacy:
whoops; let's raise this against the right repo

geomacy:
Can't figure out how to raise this against jclouds/jclouds - https://github.com/jclouds/jclouds/pulls. I will ask on the list.

geomacy:
That's a bit more like it - closing this in favour of https://github.com/jclouds/jclouds/pull/1059.

PR #16, opened 2019-02-05
switch to signaturev4

ghyster:
V2 signature is not supported across all AWS regions, V4 signature is


ghyster:
Sorry, I didn't find anything about this in the docs, I'll test with AWSS3HttpApiModule, thank you

PR #20, opened 2019-03-19
JCLOUDS-1428 - Support for SAS token based Authentication for Azure Blob Storage - Follow Up

ak58588:
Need to remove "sp" and "se" tokens from the SAS check, since there is a possible SAS format without those two tokens;
Added a unit test case.


gaul:
@nacx any feedback?

nacx:
cherry-pick to 2.1.x?

PR #21, opened 2019-04-02
Adds methods listAll and listByLocation to VirtualMachineAPI

danielestevez:
* https://docs.microsoft.com/en-us/rest/api/compute/virtualmachines/listall
* https://docs.microsoft.com/en-us/rest/api/compute/virtualmachines/listbylocation

These methods will be very useful to improve the performance of jclouds provider skipping calls to resourcegroups API (as in [here](https://github.com/danielestevez/apache-jclouds/blob/d621edd79f1d95c7bdaa4803bf36fa4d62cfdd99/providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java#L354))


providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/VirtualMachineApi.java
```diff
@@ -139,14 +151,14 @@ VirtualMachine createOrUpdate(@PathParam("vmname") String vmname,
     */
    @Named("capture")
    @POST
-   @Payload("%7B\"vhdPrefix\":\"{vhdPrefix}\",\"destinationContainerName\":\"{destinationContainerName}\",\"overwriteVhds\":\"true\"%7D")
-   @Path("/{name}/capture")
+   @Payload("%7B\"vhdPrefix\":\"{vhdPrefix}\",\"destinationContainerName\":\"{destinationContainerName}\","
+         + "\"overwriteVhds\":\"true\"%7D")
```
nacx: Put it in one line as it was before

providers/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiMockTest.java
```diff
@@ -119,6 +119,22 @@ public void testListEmpty() throws Exception {
             + "/virtualMachines?api-version=2018-06-01");
    }
 
+   public void testListAll() throws Exception {
+      server.enqueue(jsonResponse("/virtualmachinesinsubscription.json"));
+      final VirtualMachineApi vmAPI = api.getVirtualMachineApi(null);
+      assertEquals(vmAPI.listAll(), getVMListAll()); //TODO allinsubs
```
nacx: What does this TODO and the other one stand for? Can they be removed or is there anything else TBD?

nacx:
The PR is small enough to go ahead and change the adapter too.
Why isn't it changed in this PR? Having a second look I really think changing the adapter could be part of this PR. jclouds is all about the abstractions, so improving them should be the very first main focus. Let's update the adapter too in this PR.

providers/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/VirtualMachineApiMockTest.java
danielestevez: That one is already done, forgot to remove it!

danielestevez:
Well i prefer smaller PRs and these methods are just ready to be used but i can add the work i'm doing on the adapter to this same PR if it's better and merge it all when it's ready.

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/features/VirtualMachineApi.java
danielestevez: oops yes

nacx:
Yes, I think you should add your adapter work here. One mistake we've done in the past is not putting abstractions first. Everything should be driven by our abstractions, and efforts should we put around them.
The goal of this addition to the Azure APIs should be to optimize the current abstraction, not just add more methods to the API itself, so let's better add the adapter changes to this PR too. Otherwise, we continue with the tendency of adding APIs with no clear commitment to complete the abstractions later and keep growing our API footprint while keeping the abstractions incomplete.

danielestevez:
Just a quick update on the state of this PR:
The changes to the adapter were minor but i found one of those nasty ARM API behaviour with case-sensitivity.
In short, these new methods return the resourcegroup part of the resource id in capital letters in randomly and this causes some of the LiveTests to fail, so it's not safe to add them to the adapter yet
The goal to make all the provider case-insensitive (as Microsoft [recommends](https://stackoverflow.com/a/50082864/2369157) in integrations) is too wide for this PR but i found a way to fix this problem just for the resourcegroups.
Running all the LiveTests with this fix i only found a problem still in `AzureComputeServiceLiveTest#testListNodesByIds` that i will try to fix too

providers/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceLiveTest.java
```diff
@@ -67,6 +75,30 @@ public void initializeContext() {
...
+            return from.getId();
+         }
+
+      }));
+
+      final Set<? extends ComputeMetadata> listedNodes = client.listNodesByIds(nodeIds);
+
+      assertEquals(listedNodes.size(), nodes.size());
+
+      for (ComputeMetadata listedNode : listedNodes) {
```
nacx: Please add a comment explaining the justification for overriding this method to do more relaxed validations here.

danielestevez: Sorry, i just added the commits but was writing the explanation:
* First, after several tests, the case sensitivity problem will not affect the jclouds provider
* The only problem is this test, which uses the `Iterables.elementsEqual` method to compare the node list. This one makes really strict validations (entered in this [commit](https://github.com/apache/jclouds/commit/c566418d87e7b9337e57efed03a6a5e5524e690b#diff-7f014b2e6f9ed5629b02d2b1085e32f7R578) ) like the order and case sensitivity. In Azure we could have some fields (as `providerId`) changing case. (Note also than other tests like `testListNodes`, `testGet` or `testGetNodesWithDetails` actually do a much more relaxed validation)
* I'd really prefer to change the original method but i don't understand why it has all of those validations. IMHO the base method should make lesser validations and the overrode methods could add more validations if needed. 
 
To be discussed! We could

1. Override this one and add a comment
2. Change the base method

nacx: This is an Azure particular behavior and I see no point in modifying a test that works consistently across providers just for the Azure case-insensitive stuff. Let's override and add a proper comment.

nacx:
>  after several tests, the case sensitivity problem will not affect the jclouds provider

Then I understand adding the changes to the abstraction is not a big deal, right? Look forward to seeing that added to this PR.

danielestevez:
Sorry, big oversight. That commit was in another branch and i thought i had already pushed it here :sweat:

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/AzureComputeServiceAdapter.java
```diff
@@ -351,14 +351,8 @@ public void suspendNode(final String id) {
...
       ImmutableList.Builder<VirtualMachine> nodes = builder();
-      for (ResourceGroup resourceGroup : api.getResourceGroupApi().list()) {
-         List<VirtualMachine> vms = api.getVirtualMachineApi(resourceGroup.name()).list();
-         nodes.addAll(filter(vms, new Predicate<VirtualMachine>() {
-            @Override
-            public boolean apply(VirtualMachine input) {
-               return regionIds.get().isEmpty() || regionIds.get().contains(input.location());
-            }
-         }));
+      for (final String location : regionIds.get()) {
```
nacx: If `regionIds` is empty (no region constraint) this should return all nodes. It currently returns nothing.

PR #22, opened 2019-04-05
Disable PublicAccess enquiry for AzureBlob, when using SAS Auth. 

ak58588:
Since it is impossible (checked with the Azure Storage dev team) to execute getContainerACL with SAS Authorisation, here is a possible solution to omit checking the Public Access policy for Cotnainers and Blobs when using SAS, as it will fail. 

No test provided yet.


gaul:
This approach confuses me since it populates the cache with bogus data.  As I understand it, SAS authenticated requests will try to consult this `LoadingCache` which will fail since that token may not have the privilege to get container ACLs.  Instead of using fake data, instead should it throw an exception which the outer `catch (CacheLoader.InvalidCacheLoadException e)` already handles?

PR #23, opened 2019-04-10
disable AccessPolicy check for SAS Authorised azureblob

ak58588:
This is a draft. I have a problem with it: 
even though I am throwing a `CacheLoader.InvalidCacheLoadException` exception in `containerAcls` method in `AzureBlobStoreContextModule.java`, I still can't catch it in the try-catch close in `BlobPropertiesToBlobMetadata.java` with `catch (CacheLoader.InvalidCacheLoadException e)`.
Nevertheless, I can catch this` CacheLoader.InvalidCacheLoadException` when I do `catch (Exception e)`. 
Do you have any idea of the reason why this is happening and how to omit catching unspecified Exception?


nacx:
I think it probably makes more sense here to have a custom exception such as `InsufficientAccessRightsException` or something like that that properly designates the error to handle instead of a generic loading exception? /cc @gaul

ak58588:
the `catch (CacheLoader.InvalidCacheLoadException e)` can't catch the `CacheLoader.InvalidCacheLoadException`, but catches `com.google.common.util.concurrent.UncheckedExecutionException`.  getUnchecked method throws it.

providers/azureblob/src/main/java/org/jclouds/azureblob/config/InsufficientAccessRightsException.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Handles the inability of SAS Authentication string to authenticate the getAcl and setAcl requests.
+ * 
+ */
+ 
+package org.jclouds.azureblob.config;
+
+import com.google.common.util.concurrent.UncheckedExecutionException;
+
+public class InsufficientAccessRightsException extends UncheckedExecutionException {
```
nacx: I think this should just extend Exception or RuntimeException, depending on whether you want to make this checked or not.

providers/azureblob/src/main/java/org/jclouds/azureblob/blobstore/functions/BlobPropertiesToBlobMetadata.java
```diff
@@ -58,6 +59,8 @@ public MutableBlobMetadata apply(BlobProperties from) {
             PublicAccess containerAcl = containerAcls.getUnchecked(from.getContainer());
             if (containerAcl != PublicAccess.PRIVATE)
                to.setPublicUri(from.getUrl());
+         } catch (UncheckedExecutionException ue) {
+               to.setPublicUri(from.getUrl());
```
nacx: Would something like the following work?
```java
} catch (Exception ex) {
   InsufficientAccessRightsException iare = Throwables2.getFirstThrowableOfType(ex, InsufficientAccessRightsException.class);
   if (iare == null) throw ex; // Propagate any other uncaught exceptions
   to.setPublicUri(from.getUrl()); // handle our own exception gracefully
}
```

gaul: Should this call `setPublicUri` when catching the exception?

apis/openstack-keystone/pom.xml
```diff
@@ -108,6 +108,7 @@
     <!-- Disabling error-prone compiler due to: https://github.com/google/error-prone/issues/711
          The fix is only available in error-prone versions that do not support Java 7 -->
       <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
```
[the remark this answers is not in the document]
ak58588: I guess it corresponds to your master branch - I had created this draft pull request before this has been pushed

providers/azureblob/src/main/java/org/jclouds/azureblob/config/InsufficientAccessRightsException.java
ak58588: implemented both: extending RuntimeException in the 'staging' branch and Exception in the 'test' branch

providers/azureblob/src/main/java/org/jclouds/azureblob/blobstore/functions/BlobPropertiesToBlobMetadata.java
nacx: I'm not that familiar with the contract for this API and try/catch block. So... Sure, let's call `setPublicUri` wherever it makes sense. I just tried to suggest a way of capturing the _real_ exception we care about.
Once we capture the exception we want, let's make sure we set the URI when needed. I'll defer that to you, as you know the API better :)

gaul: I agree with your comment!  But I am suggesting that if the bucket is non-public, it should not have a public URI.

ak58588: Shall we then omit setting the URI at all? What happens if this parameter is unset?

ak58588: hi @gaul! Could you please take a look at the code now? I have implemented the changes, proposed above

ak58588:
hi @gaul, could you please review this?

providers/azureblob/src/main/java/org/jclouds/azureblob/blobstore/functions/BlobPropertiesToBlobMetadata.java
```diff
@@ -58,8 +61,9 @@ public MutableBlobMetadata apply(BlobProperties from) {
             PublicAccess containerAcl = containerAcls.getUnchecked(from.getContainer());
             if (containerAcl != PublicAccess.PRIVATE)
                to.setPublicUri(from.getUrl());
-         } catch (CacheLoader.InvalidCacheLoadException e) {
-            // nulls not permitted from cache loader
+         } catch (Exception ex) {
+            InsufficientAccessRightsException iare = Throwables2.getFirstThrowableOfType(ex, InsufficientAccessRightsException.class);
+            if (iare == null) throw ex;
```
gaul: Maybe add a comment explaining that this is not a publicly accessible object?

gaul:
Also please remove draft metadata.

providers/azureblob/src/main/java/org/jclouds/azureblob/blobstore/functions/BlobPropertiesToBlobMetadata.java
ak58588: Done

ak58588:
Unfortunately the Jenkins jobs keep failing, this may be because many things have been changed in the background while the PR was existing, such as checkstyle-suppressions.xml, which are the reason why the Jenkins jobs keep failing. I will create a new PR and commit my changes there

PR #24, opened 2019-04-10
JCLOUDS-1495: maven plugins are not correctly referred to

oflebbe:
Some maven plugins are not properly addressed.


nacx:
Thanks!

PR #27, opened 2019-04-30
JCLOUDS-1497: Fix checkstyle-suppressions for jcloud-labs

oflebbe:
This is the jclouds part of fixing checkstyle. jclouds-labs has a couple of violations as well

### DIFF
--- a/resources/checkstyle.xml
+++ b/resources/checkstyle.xml
@@ -48,7 +48,7 @@
         </module>
         <module name="NoWhitespaceBefore"/>
         <module name="RedundantImport"/>
-        <module name="RedundantModifier"/>
+       <!-- <module name="RedundantModifier"/> -->
         <module name="SimplifyBooleanExpression"/>
         <module name="TypeName"/>
         <module name="UnnecessaryParentheses"/>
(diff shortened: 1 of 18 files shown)


nacx:
Closing and re-opening to trigger the build again

resources/checkstyle.xml
```diff
@@ -48,7 +48,7 @@
         </module>
         <module name="NoWhitespaceBefore"/>
         <module name="RedundantImport"/>
-        <module name="RedundantModifier"/>
+       <!-- <module name="RedundantModifier"/> -->
```
nacx: The PR LGTM, but why are you removing this check?

oflebbe: We have some static imports (search for "import static") in the code. I am doubting it is worth the effort to fix it. Wdyt ?

nacx: I'm wondering why this check was not failing before. Anyway, I agree there is little value in this one.

gaul: This check should find redundant modifiers like `public` in interfaces, not static imports.

nacx: Right. I ran a local build and only in `core` there were more than 40 violations. I fixed them, then found the same in another project... So I agree the value this check provides is close to zero and I think removing it is OK.

oflebbe: I run the report again, and yeah sorry you are right: The violations where way down the file not on top where the "import static" was used.

PR #29, opened 2019-05-06
JCLOUDS-1500: Update gson to 2.8.5

oflebbe:
If we do not have to consider to fix JCLOUDS-1166 in the first place (using of internal gson API) , updating gson is a piece of cake

### DIFF
--- a/core/src/main/java/org/jclouds/json/internal/DeserializationConstructorAndReflectiveTypeAdapterFactory.java
+++ b/core/src/main/java/org/jclouds/json/internal/DeserializationConstructorAndReflectiveTypeAdapterFactory.java
@@ -26,6 +26,7 @@ import java.lang.reflect.Modifier;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


nictas:
We really need this, since we're also using https://mvnrepository.com/artifact/org.immutables/gson/2.7.5 in our project, but it depends on Gson 2.8.5. Is there any timeline for when this is going to be released?

nacx:
@nictas we're actively discussing this in [JCLOUDS-1166](https://issues.apache.org/jira/browse/JCLOUDS-1166). Your feedback would be valuable too, so please, feel free to comment on the issue.

nacx:
I've rebased and re-applied your change on top of the latest master now that #35 is merged. I hope I'm not breaking the build :)
Mind testing if this change is still OK and fixes the issue you were having?
Hopefully, we can move forward now and upgrade this and other dependencies that were preventing us from upgrading to newer versions of the JDK.

demobox:
> I've rebased and re-applied your change on top of the latest master now that #35 is merged

Yay - thanks, Ignasi!

PR #30, opened 2019-05-14
Checks provisioning state in Rule resource instead of Group

danielestevez:
* We were checking the provisioningState security group resource. When updating rules concurrently this check returns OK (the security group is ok) but the actual rules are still being updated (created/deleted) causing random problems.

* PR implements provisioningState in `NetworkSecurityRuleProperties` to do this check properly (https://docs.microsoft.com/en-us/rest/api/virtualnetwork/securityrules/get#securityrule)

### DIFF
--- a/providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeSecurityGroupExtension.java
+++ b/providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeSecurityGroupExtension.java
@@ -294,11 +294,11 @@ public class AzureComputeSecurityGroupExtension implements SecurityGroupExtensio
 
       for (NetworkSecurityRule matchingRule : rules) {
          logger.debug(">> deleting network security rule %s from %s...", matchingRule.name(), group.getName());
-         ruleApi.delete(matchingRule.name());
-         URI uri = api.getNetworkSecurityRuleApi(resourceGroupAndName.resourceGroup(), networkSecurityGroup.name()).delete(resourceGroupAndName.name());
+         URI uri = ruleApi.delete(matchingRule.name());
          if (uri != null) {
-            resourceDeleted.apply(uri);
+            checkState(resourceDeleted.apply(uri), "Rule %s could not be deleted", matchingRule.id());
          }
+
       }
 
       return getSecurityGroupById(group.getId());
--- a/providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeSecurityGroupExtension.java
+++ b/providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeSecurityGroupExtension.java
@@ -30,13 +30,13 @@ import java.net.URI;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
-
 import javax.annotation.Resource;
 import javax.inject.Inject;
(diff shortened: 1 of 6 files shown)


providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeSecurityGroupExtension.java
```diff
@@ -295,9 +295,10 @@ public boolean apply(NetworkSecurityRule input) {
       for (NetworkSecurityRule matchingRule : rules) {
          logger.debug(">> deleting network security rule %s from %s...", matchingRule.name(), group.getName());
          ruleApi.delete(matchingRule.name());
-         checkState(
-               securityGroupAvailable.create(resourceGroupAndName.resourceGroup()).apply(networkSecurityGroup.name()),
-               "Security group was not updated in the configured timeout");
+         URI uri = api.getNetworkSecurityRuleApi(resourceGroupAndName.resourceGroup(), networkSecurityGroup.name()).delete(resourceGroupAndName.name());
```
nacx: This looks wrong:
* The rule is already being deleted in line 297.
* You are passing the security group name to the delete method, not the rule name.

Remove this line, capture the URL in line 297 and check on that.

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeSecurityGroupExtension.java
```diff
@@ -295,9 +295,10 @@ public boolean apply(NetworkSecurityRule input) {
...
+         if (uri != null) {
+            resourceDeleted.apply(uri);
```
nacx: Let's capture the result and verify with `checkState` to keep the current behavior and feedback.

danielestevez:
Thanks @nacx for noticing. Tests however passed but code was clearly wrong

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/extensions/AzureComputeSecurityGroupExtension.java
```diff
@@ -294,10 +294,11 @@ public boolean apply(NetworkSecurityRule input) {
 
       for (NetworkSecurityRule matchingRule : rules) {
          logger.debug(">> deleting network security rule %s from %s...", matchingRule.name(), group.getName());
-         ruleApi.delete(matchingRule.name());
-         checkState(
-               securityGroupAvailable.create(resourceGroupAndName.resourceGroup()).apply(networkSecurityGroup.name()),
-               "Security group was not updated in the configured timeout");
+         URI uri = ruleApi.delete(matchingRule.name());
+         if (uri != null) {
+            checkState(resourceDeleted.apply(uri), "Rule %s could not be deleted", matchingRule.id());
```
nacx: You are just waiting for a certain period, so you don't actually know if the rule is gonna be deleted in the end
```suggestion
            checkState(resourceDeleted.apply(uri), "Rule %s was not be deleted in the configured timeout", matchingRule.id());
```

danielestevez: Right! Modified

danielestevez:
Thanks for merging so quickly! I was thinking we could include this one in 2.1.x? 
Maybe i should have opened the PR against such branch better

PR #33, opened 2019-05-17
Implements listAll operation for VirtualNetworkAPI

danielestevez:
This should be the last of these improvements.

Again jclouds doesn't use the method


nacx:
LGTM but tests need to be fixed

danielestevez:
sorry, dumb mistake there

danielestevez:
Found a bug in method signature. resourcegroup param should be nullable. Fixing....

danielestevez:
And ready now!

PR #34, opened 2019-05-27
JCLOUDS-1428 - Support for SAS token based Authentication for Azure Blob Storage

ak58588:
Removed the ACL check for the AzureBlobs, authorised with SAS Auth String. 
Implemented all change requests from the previous PR #23.


gaul:
Pushed to master and 2.1.x.  Thank you for your contribution @ak58588!

ak58588:
Thanks @gaul and @nacx !

nacx:
Thanks for your patches!

PR #37, opened 2019-06-15
Fix regression in azureblob and properly catch unchecked exceptions

nacx:
Fix a regression introduced in https://github.com/apache/jclouds/pull/34

The method implements `Function` and cannot declare an exception to be thrown, so we should only propagate RuntimeExceptions.
It is fine to just catch that since the methods inside the catch block are not meant to throw checked exceptions.

/cc @ak58588


gaul:
Why does 2.1.x need this but not master?

nacx:
I'm not sure, but definitely makes sense to post to master too. Don't know why it isn't failing there.

PR #38, opened 2019-06-17
Add OSGi exports to jclouds-gson

nacx:


nacx:
After a closer look, it seems like the renamed packages (`org.jclouds.json.gson.internal`) does not appear in the generated manifest exports.
It could be potentially due to the mentioned issue and incompatibility with the shade plugin and the Maven reactor, so the bundle plugin does not _see_ the relocated package.

I'm a bit out of cycles this week, and would highly appreciate if someone could have a look at this?

nacx:
Ok, it is a bit ugly, but after quite some research this is the only working way I've found to make the shade and bundle plugins work together. Basically, we need to split the work to enforce order and make sure the bundle plugin sees all the classes referenced in the exports.

gson/pom.xml
```diff
@@ -26,56 +26,21 @@
...
-  <!-- 
-    JCLOUDS-1166: This project relocates the 'com.google.gson.internal' package
-    so we can keep using its contents in OSGi environments.
-    This is a temporal solution and we should try to remove the usages of the classes in
-    there and remove this once that is done.
-  -->
-  <artifactId>jclouds-gson</artifactId>
-  <name>jclouds gson copy without internal packages</name>
-  <packaging>bundle</packaging>
+  <artifactId>jclouds-gson-modules</artifactId>
```
demobox: [minor] Just curious what this aggregator POM does? I guess it's because we didn't want two gson-related bundles at the top level (if so, definitely agree that that)?

gson/gson-bundle/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+  <properties>
+    <jclouds.osgi.export>
+      org.jclouds.json.gson.internal*;version=${project.version};-noimport:=true,
+      !com.google.gson.internal*,
+      com.google.gson*;version=${project.version};-noimport:=true,
+    </jclouds.osgi.export>
+    <jclouds.osgi.import>*</jclouds.osgi.import>
+  </properties>
+
+  <dependencies>
```
demobox: Argh - this is what you meant by having to split the shading and the bundling up. Grr. I guess it sort-of makes sense since the shading has to happen before the bundle manifest generation.

Out of curiosity, did you try binding the shade plugin to a phase _before_ `package`, e.g. [`prepare-package`](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#Lifecycle_Reference)? Definitely not important if this ends up working, I should add!

nacx: Yep, I tried that and did not work, also binding the bundle plugin after shade explicitly, same thing.

demobox: _sigh_. Thanks for explaining!

PR #39, opened 2019-06-17
Plugin config in pluginManagement for proper reuse in child projects

nacx:
I think this should fix the current jclouds-labs issue that is failing because apparently, it's not picking the proper config from the duplicate class finder plugin.


nacx:
Yes, so for some reason the labs build is complaining about duplicate deps, when they are explicitly configured in the plugin.

This highlights that config is not being inherited and that's the reason why probably we have a copy of the rat plugin config in labs.

This change should fix config inheritance for rat and the duplicate finder (which should be used by every module) and labs build should be back to green. We'll be able to cleanup the other repos poms too after this.

PR #41, opened 2019-08-13
JCLOUDS-1510: ParseSax use charset constant for all String creations

roded:
This commit prevents the `convertStreamToStringAndParse` method from
failing XML validation on environments in which the default charset is
not compatible with UTF-8.

### DIFF
--- a/core/src/main/java/org/jclouds/http/functions/ParseSax.java
+++ b/core/src/main/java/org/jclouds/http/functions/ParseSax.java
@@ -25,6 +25,7 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.StringReader;
+import java.nio.charset.StandardCharsets;
 
 import javax.annotation.Resource;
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/jclouds/http/functions/ParseSax.java
```diff
@@ -93,7 +94,7 @@ private T convertStreamToStringAndParse(HttpResponse response) {
       String from = null;
       try {
          byte[] fromBytes = closeClientButKeepContentStream(response);
-         from = new String(fromBytes);
+         from = new String(fromBytes, StandardCharsets.UTF_8);
```
gaul: I think this would be more correct to use the charset from the Content-Type header.  However I doubt any provider use charsets other than UTF-8 or ASCII?

roded: Probably not, but I can't say for sure.
I can get the charset from the Content-Type header (if it exists there) - are there some HTTP utilities already in the classpath I can use to avoid parsing the Content-Type string myself?

gaul:
Pushed to master as dfeafcc6fb914a6e5aa5293fcdf188e39a87842a and 2.1.x as d4b9f03059fdd391bcca243b0abf3bc37243e8b8.  Thank you for your contribution @roded!

PR #43, opened 2019-08-26
JCLOUDS-1511: allow configuration of S3 to use V4 signatures

roded:
Added the "jclouds.v4signatures" property to allow configuring S3 to
use V4 signatures instead of the default V2.

This is a draft. Comments are more than welcome (possible issues: property name, logic in a Guice module). Specifically, I'm not sure where testing of this functionality should go - if there's an example to follow it would be helpful.

### DIFF
--- a/apis/s3/src/main/java/org/jclouds/s3/S3ApiMetadata.java
+++ b/apis/s3/src/main/java/org/jclouds/s3/S3ApiMetadata.java
@@ -18,7 +18,6 @@ package org.jclouds.s3;
 
 import static org.jclouds.Constants.PROPERTY_IDEMPOTENT_METHODS;
 import static org.jclouds.Constants.PROPERTY_RELAX_HOSTNAME;
-import static org.jclouds.Constants.PROPERTY_V4_REQUEST_SIGNATURES;
 import static org.jclouds.aws.reference.AWSConstants.PROPERTY_AUTH_TAG;
 import static org.jclouds.aws.reference.AWSConstants.PROPERTY_HEADER_TAG;
 import static org.jclouds.blobstore.reference.BlobStoreConstants.PROPERTY_BLOBSTORE_DIRECTORY_SUFFIX;
@@ -26,6 +25,7 @@ import static org.jclouds.blobstore.reference.BlobStoreConstants.PROPERTY_USER_M
 import static org.jclouds.reflect.Reflection2.typeToken;
 import static org.jclouds.s3.reference.S3Constants.PROPERTY_JCLOUDS_S3_CHUNKED_SIZE;
 import static org.jclouds.s3.reference.S3Constants.PROPERTY_S3_SERVICE_PATH;
+import static org.jclouds.s3.reference.S3Constants.PROPERTY_S3_V4_REQUEST_SIGNATURES;
 import static org.jclouds.s3.reference.S3Constants.PROPERTY_S3_VIRTUAL_HOST_BUCKETS;
 
 import java.net.URI;
@@ -80,7 +80,7 @@ public class S3ApiMetadata extends BaseHttpApiMetadata {
       properties.setProperty(PROPERTY_S3_VIRTUAL_HOST_BUCKETS, "false");
       properties.setProperty(PROPERTY_RELAX_HOSTNAME, "true");
       properties.setProperty(PROPERTY_BLOBSTORE_DIRECTORY_SUFFIX, "/");
-      properties.setProperty(PROPERTY_V4_REQUEST_SIGNATURES, "false");
+      properties.setProperty(PROPERTY_S3_V4_REQUEST_SIGNATURES, "false");
       properties.setProperty(PROPERTY_USER_METADATA_PREFIX, String.format("x-${%s}-meta-", PROPERTY_HEADER_TAG));
       properties.setProperty(PROPERTY_IDEMPOTENT_METHODS, "DELETE,GET,HEAD,OPTIONS,POST,PUT");
 
--- a/apis/s3/src/main/java/org/jclouds/s3/S3ApiMetadata.java
+++ b/apis/s3/src/main/java/org/jclouds/s3/S3ApiMetadata.java
@@ -18,6 +18,7 @@ package org.jclouds.s3;
 
 import static org.jclouds.Constants.PROPERTY_IDEMPOTENT_METHODS;
 import static org.jclouds.Constants.PROPERTY_RELAX_HOSTNAME;
+import static org.jclouds.Constants.PROPERTY_V4_REQUEST_SIGNATURES;
 import static org.jclouds.aws.reference.AWSConstants.PROPERTY_AUTH_TAG;
 import static org.jclouds.aws.reference.AWSConstants.PROPERTY_HEADER_TAG;
 import static org.jclouds.blobstore.reference.BlobStoreConstants.PROPERTY_BLOBSTORE_DIRECTORY_SUFFIX;
@@ -79,6 +80,7 @@ public class S3ApiMetadata extends BaseHttpApiMetadata {
       properties.setProperty(PROPERTY_S3_VIRTUAL_HOST_BUCKETS, "false");
       properties.setProperty(PROPERTY_RELAX_HOSTNAME, "true");
       properties.setProperty(PROPERTY_BLOBSTORE_DIRECTORY_SUFFIX, "/");
+      properties.setProperty(PROPERTY_V4_REQUEST_SIGNATURES, "false");
       properties.setProperty(PROPERTY_USER_METADATA_PREFIX, String.format("x-${%s}-meta-", PROPERTY_HEADER_TAG));
       properties.setProperty(PROPERTY_IDEMPOTENT_METHODS, "DELETE,GET,HEAD,OPTIONS,POST,PUT");
 
--- a/apis/s3/src/main/java/org/jclouds/s3/config/S3HttpApiModule.java
+++ b/apis/s3/src/main/java/org/jclouds/s3/config/S3HttpApiModule.java
@@ -24,6 +24,7 @@ import java.util.concurrent.TimeUnit;
 import javax.inject.Named;
 import javax.inject.Singleton;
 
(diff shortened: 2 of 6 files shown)


apis/s3/src/main/java/org/jclouds/s3/config/S3HttpApiModule.java
```diff
@@ -183,8 +183,16 @@ protected void bindErrorHandlers() {
...
+   @Provides
+   @Singleton
+   protected void bindRequestSigner(@Named(Constants.PROPERTY_V4_REQUEST_SIGNATURES) boolean v4Signatures) {
+      Class<? extends RequestAuthorizeSignature> clazz;
+      if (v4Signatures) {
+         clazz = RequestAuthorizeSignatureV4.class;
+      } else {
+         clazz = RequestAuthorizeSignatureV2.class;
+      }
+      bind(RequestAuthorizeSignature.class).to(clazz).in(Scopes.SINGLETON);
```
nacx: Hmm, does this actually work? I don't know if `bind` statements can fo in a provider method, but it looks semantically weird. Provider methods are supposed to return objects that can be injected. Instead, I'd change this method to something like this (I haven't tried it, but you'll get the idea):

```java
@Provides
@Singleton
protected RequestAuthorizeSignature bindRequestSigner(
      @Named(Constants.PROPERTY_V4_REQUEST_SIGNATURES) boolean v4Signatures,
      RequestAuthorizeSignatureV2 v2,
      RequestAuthorizeSignatureV4 v4) {
   return v4Signatures ? v4 : v2;
}
```

The idea is to have the provider method get injected the possible instances and return the right one when an injection is requested.

Adding a test for this should be pretty straightforward. I'd just create a test class that creates a raw Guice injector with this module and the value of the property, then call `injector.getInstance` and verify that the returned object is of the expected type.

nacx: A better approach to avoid instantiating the objects you're not using would be something like:

```java
@Provides
@Singleton
protected RequestAuthorizeSignature bindRequestSigner(
      @Named(Constants.PROPERTY_V4_REQUEST_SIGNATURES) boolean v4Signatures,
      Injector i) {
   return i.getInstance(v4Signatures ? RequestAuthorizeSignatureV4.class : RequestAuthorizeSignatureV2.class);
}
```

roded: Thanks for the review @nacx . You're right about the `bind` in a provider method - I have not run this yet. Will do so in a unit test.

apis/s3/src/main/java/org/jclouds/s3/S3ApiMetadata.java
```diff
@@ -79,6 +80,7 @@ public static Properties defaultProperties() {
       properties.setProperty(PROPERTY_S3_VIRTUAL_HOST_BUCKETS, "false");
       properties.setProperty(PROPERTY_RELAX_HOSTNAME, "true");
       properties.setProperty(PROPERTY_BLOBSTORE_DIRECTORY_SUFFIX, "/");
+      properties.setProperty(PROPERTY_V4_REQUEST_SIGNATURES, "false");
```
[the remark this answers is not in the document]
roded: If so, I could be able to remove the binding of `RequestAuthorizeSignature` from `AWSS3HttpApiModule`.

apis/s3/src/main/java/org/jclouds/s3/config/S3HttpApiModule.java
roded: > Adding a test for this should be pretty straightforward. I'd just create a test class that creates a raw Guice injector with this module and the value of the property, then call `injector.getInstance` and verify that the returned object is of the expected type.

Is there a recommended way to handle the missing dependencies of this module in the test? I gather that this module requires quite a few dependencies from other modules which do not have mocks. 

I suppose I could extract the provider of the `RequestAuthorizeSignature` to its own module, that way its testing should be simpler.

apis/s3/src/test/java/org/jclouds/s3/config/S3HttpApiModuleTest.java
```diff
@@ -0,0 +1,63 @@
+package org.jclouds.s3.config;
```
nacx: Add the license header

apis/s3/src/test/java/org/jclouds/s3/config/S3HttpApiModuleTest.java
```diff
@@ -0,0 +1,63 @@
...
+        List<Module> modules = ImmutableList.<Module>of(new BaseRestApiTest.MockModule(),
+                new CredentialStoreModule(new ConcurrentHashMap<String, ByteSource>()),
+                new S3BlobStoreContextModule(),
+                new NullLoggingModule(),
+                new S3HttpApiModule<>(),
+                new LocationModule(),
+                abstractModule);
+        return Guice.createInjector(modules).getInstance(RequestAuthorizeSignature.class);
+    }
+}
```
nacx: Nice! :)

nacx:
@gaul @demobox do you know why CI builds are not being triggered?

gaul:
No idea -- this worked as recently as a week ago and CI does not show any configuration changes.

apis/s3/src/test/java/org/jclouds/s3/config/S3HttpApiModuleTest.java
roded: This does not yet work. Other dependencies are missing:
`com.google.inject.CreationException: Guice creation errors:

1) No implementation for com.google.common.base.Supplier<org.jclouds.domain.Credentials> annotated with @org.jclouds.location.Provider() was bound.
  while locating com.google.common.base.Supplier<org.jclouds.domain.Credentials> annotated with @org.jclouds.location.Provider()
    for parameter 1 at org.jclouds.rest.internal.RestAnnotationProcessor.<init>(RestAnnotationProcessor.java:137)
  while locating org.jclouds.rest.internal.RestAnnotationProcessor
    for parameter 0 at org.jclouds.s3.blobstore.S3BlobRequestSigner.<init>(S3BlobRequestSigner.java:61)
  at org.jclouds.s3.blobstore.config.S3BlobStoreContextModule.bindRequestSigner(S3BlobStoreContextModule.java:55)
...
`

apis/s3/src/main/java/org/jclouds/s3/config/S3HttpApiModule.java
```diff
@@ -183,8 +184,11 @@ protected void bindErrorHandlers() {
       bind(HttpErrorHandler.class).annotatedWith(ServerError.class).to(ParseS3ErrorFromXmlContent.class);
    }
 
-   protected void bindRequestSigner() {
-      bind(RequestAuthorizeSignature.class).to(RequestAuthorizeSignatureV2.class).in(Scopes.SINGLETON);
+   @Provides
+   @Singleton
+   protected RequestAuthorizeSignature bindRequestSigner(
+           @Named(S3Constants.PROPERTY_S3_V4_REQUEST_SIGNATURES) boolean v4Signatures, Injector injector) {
+      return injector.getInstance(v4Signatures ? RequestAuthorizeSignatureV4.class : RequestAuthorizeSignatureV2.class);
```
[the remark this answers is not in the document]
roded: Could you explain your comment a bit for my sake?
Will `AWSS3BlobRequestSignerV4` not use the `RequestAuthorizeSignature` bound by `S3HttpApiModule`?

demobox:
> do you know why CI builds are not being triggered?

Looks like it's working on #44 

image

Will close and reopen to see if that changes anything.

demobox:
> reopened this 13 seconds ago

Hrm, no CI yet. Perhaps once the requested change is made? Otherwise, we could try proposing the change again as a new PR?

apis/s3/src/main/java/org/jclouds/s3/config/S3HttpApiModule.java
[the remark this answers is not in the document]
gaul: I believe that there are two signing mechanisms: one for general URLs, and one for pre-signed URLs which the application server can give to partially-trusted clients.  For consistency we should use the same signer for both.

apis/s3/src/test/java/org/jclouds/s3/config/S3HttpApiModuleTest.java
nacx: Apologies for the late reply. I found it hard to have cycles to get back to this.
You can use the following code instead (taken mostly from the signature fitlers test classes):

```java
private RequestAuthorizeSignature getRequestAuthorizeSignature(final String s3V4Value) {
   Properties overrides = new Properties();
   overrides.setProperty(PROPERTY_S3_V4_REQUEST_SIGNATURES, s3V4Value);
   
   Injector injector = ContextBuilder.newBuilder(new S3ApiMetadata())
         .credentials("foo", "bar")
         .overrides(overrides)
         .modules(ImmutableList.<Module> of(
               new BaseRestApiTest.MockModule(), 
               new NullLoggingModule(),
               new TestS3HttpApiModule()))
         .buildInjector();
   
   return injector.getInstance(RequestAuthorizeSignature.class);
}

@ConfiguresHttpApi
private static final class TestS3HttpApiModule extends S3HttpApiModule<S3Client> {
   private static final String TIMESTAMP = "Thu, 03 Feb 2015 07:11:11 GMT";
   
   @Override
   protected String provideTimeStamp(@TimeStamp Supplier<String> cache) {
      return TIMESTAMP;
   }

   @Override
   protected Supplier<Date> provideTimeStampCacheDate(@Named(Constants.PROPERTY_SESSION_INTERVAL) long seconds,
         @TimeStamp final Supplier<String> timestamp, final DateService dateService) {
      return Suppliers.ofInstance(dateService.rfc822DateParse(TIMESTAMP));
   }
}
```

roded:
I'll close this PR as I don't think I'll be able to reattempt it anytime soon.

PR #44, opened 2019-09-01
Update README copyright dates

demobox:
Also opening this to test pull request CI triggers

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,6 @@ For more information about using or contributing to jclouds, please visit our we
 
 License
 -------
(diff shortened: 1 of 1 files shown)


nacx:
Asked infra and Jenkins PR builds are only triggered for PRs from trusted members (PMC, Committers) but not for every PR.
We want to be able to build every PR, so I've just filed an issue to get TravisCI enabled in our repos, so we can have the PR builds back: https://issues.apache.org/jira/browse/INFRA-19043

demobox:
> nacx approved these changes 9 hours ago

Thanks, Ignasi!

PR #45, opened 2019-09-16
Add azure disk sku

pimuzzo:
Fix https://issues.apache.org/jira/browse/JCLOUDS-1514


providers/azurecompute-arm/src/test/java/org/jclouds/azurecompute/arm/features/DiskApiMockTest.java
```diff
@@ -78,6 +79,27 @@ public void createDiskWithTags() throws InterruptedException {
...
 
+   public void createDiskWithTagsAndSku() throws InterruptedException {
+
+      server.enqueue(jsonResponse("/creatediskwithtagsandskuresponse.json").setResponseCode(200));
+
+      final DiskApi diskApi = api.getDiskApi(resourcegroup);
+
+      DiskProperties properties = DiskProperties.builder().diskSizeGB(2).creationData(CreationData.create(CreationData.CreateOptions.EMPTY)).build();
+
+      DiskSku sku = DiskSku.builder().name("Premium_LRS").build();
```
nacx: Is there an enum of the existing valid Disk types?

pimuzzo: According to azure docs only 4 values are allowed, I'll add an enum

providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/DiskStorageAccountType.java
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+package org.jclouds.azurecompute.arm.domain;
+
+/**
+ * DiskStorageAccountTypes used in Azure.
+ *
+ * @see <a href="https://docs.microsoft.com/en-us/rest/api/compute/disks/createorupdate#diskstorageaccounttypes">this page</a>
+ */
+public enum DiskStorageAccountType
```
nacx: I've just seen we already had the [StorageAccountType](https://github.com/apache/jclouds/blob/master/providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/domain/StorageAccountType.java) for disks. Can you add the missing values there instead? (Apologies for the noise, it's been quite some time since the last time I used the ARM provider :)).

PR #46, opened 2019-09-18
Add OS disk type in template options

pimuzzo:
Fix https://issues.apache.org/jira/browse/JCLOUDS-1514


providers/azurecompute-arm/src/main/java/org/jclouds/azurecompute/arm/compute/options/AzureTemplateOptions.java
```diff
@@ -138,6 +140,12 @@ public AzureTemplateOptions customData(String customData) {
       return this;
    }
 
+   public AzureTemplateOptions osDiskStorageType(StorageAccountType osDiskStorageType)
```
nacx: [style nit] please put the `{` in the same line as the method declaration
```suggestion
   public AzureTemplateOptions osDiskStorageType(StorageAccountType osDiskStorageType) {
```

PR #48, opened 2019-10-10
Fix null content-length header on HEAD requests

xavierb-amadeus:
proposal fix for issue reported in
https://issues.apache.org/jira/projects/JCLOUDS/issues/JCLOUDS-1505


gaul:
Also pushed to 2.1.x as 6303d9630840e99ea150e6ad192ed4c8b89b8295.  Thank you for your contribution @xavierb-amadeus!

PR #49, opened 2019-10-13
JCLOUDS-1520: change UntrustedSSLContextSupplier to return the same S…

roded:
…SLContext

Using the same SSLContext prevents consistent cache misses on the JVM's KeepAliveCache
when attempting to reuse TLS connections.

### DIFF
--- a/core/src/main/java/org/jclouds/http/config/SSLModule.java
+++ b/core/src/main/java/org/jclouds/http/config/SSLModule.java
@@ -74,24 +74,22 @@ public class SSLModule extends AbstractModule {
 
    @Singleton
    public static class UntrustedSSLContextSupplier implements Supplier<SSLContext> {
-      private final TrustAllCerts trustAllCerts;
+      private final SSLContext sslContext;
 
       @Inject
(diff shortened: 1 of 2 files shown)


nacx:
Thanks! Mind rebasing to the latest master so we have th CI build feedback here?

roded:
Travis build failed due to timeout. I've noticed the same issue locally at one point. Not sure how to get around it though.

nacx:
Did it fail locally at the same point?

roded:
Yeah, tests in org.jclouds.encryption.bouncycastle seem to have an inconsistent freezing issue. We've experienced this previously on two different Linux developer machines but haven't figured it out the reason.

roded:
@nacx can this be merged?

nacx:
Sure. Take into account, though, that the 2.2.0 and 2.1.3 releases were cut before this PR was opened. They are under vote and are likely to be released on Thursday. JFYI to avoid any confusion, even when we announce the releases tomorrow and this has been merged today, this fix will not be part of 2.2.0 or 2.1.3.

roded:
Got it, thanks for the clarification.

PR #51, opened 2019-10-22
Fix infinite socket timeout when using ApacheHCHttpCommandExecutorServiceModule

xavierb-amadeus:
Because the Java type BasicHttpParams wasn't matching the one returned
by provider "newBasicHttpParams" (which is returning a HttpParams type), it was configuring the Apache HTTP client with default settings (including connection and socket timeout), ignoring the ones of JClouds which were must safer in this regard (JCloudsones are currently 60sec for both, while Apache HC ones are -1, so infinite). It also prevented to override those default/infinite timeout settings with JClouds ones.


nacx:
Cherry-picked to `2.2.x`

xavierb-amadeus:
Hi JClouds maintainers,

This commit doesn't seem to be part of the latest 2.2.0 JClouds release. Any idea if that was intentional and if so what's the planned JClouds version to have this fixed released with ? 

Many thanks in advance for your reply.
Cheers,
Xavier

nacx:
I guess because the PR was opened on Oct 22nd and jclouds 2.2.0  was released the day before.

PR #54, opened 2019-11-18
Downgrade BouncyCastle to 1.60, which works with Oracle JDK 1.7

coheigea:
This version fixes the CVEs, yet also works with Oracle JDK 1.7.

@nacx


nacx:
Thanks, @coheigea! 🙇‍♂️
I think we could also consider moving the default build to JDK 8. It would just be a matter of upgrading the `jdk.version` property in the parent `pom.xml`, but I recall some issues around terribly slow builds due to a JDK issue with reflection in Java 8.
Just mentioning in case you wanna give it a try! :)

PR #56, opened 2019-11-27
JCLOUDS-1529: Do not export org.jclouds.json.gson.internal because it causes a

lemmy:
use constraint violation with OSGi.

https://issues.apache.org/jira/browse/JCLOUDS-1529

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -39,7 +39,11 @@
 
   <properties>
     <jclouds.osgi.import>*</jclouds.osgi.import>
-    <jclouds.osgi.export>org.jclouds*;version=${project.version};-noimport:=true</jclouds.osgi.export>
+    <!-- https://issues.apache.org/jira/browse/JCLOUDS-1529 -->
+    <jclouds.osgi.export>
+      !org.jclouds.json.gson.internal*,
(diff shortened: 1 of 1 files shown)


core/pom.xml
```diff
@@ -39,7 +39,11 @@
 
   <properties>
     <jclouds.osgi.import>*</jclouds.osgi.import>
-    <jclouds.osgi.export>org.jclouds*;version=${project.version};-noimport:=true</jclouds.osgi.export>
+    <!-- https://issues.apache.org/jira/browse/JCLOUDS-1529 -->
+    <jclouds.osgi.export>
+      !org.jclouds.json.gson.internal*,
```
nacx: [nit] Please add a comment that this package is already exported by `jclouds-gson`.

nacx:
Cherry-picked to 2.2.x too

nacx:
FTR, snapshots with the fix will be available as soon as the builds finish:
https://builds.apache.org/view/J/view/jclouds/job/jclouds/
https://builds.apache.org/view/J/view/jclouds/job/2.2.x/job/jclouds-2.2.x/

PR #58, opened 2020-01-16
Fix BlobMetadata null size when using ApacheHCHttpCommandExecutorServ…

xavierb-amadeus:
…iceModule

JClouds is apparently exclusively using the Payload object from the HTTP
response to fill in the size of the BlobMetadata (when calling
blobStore.blobMetadata(...) ) - adapt this driver accordingly otherwise
we systematically get null size BlobMetadata out of it.


xavierb-amadeus:
replaced by https://github.com/apache/jclouds/pull/59

PR #59, opened 2020-01-16
Fix BlobMetadata null size when using ApacheHCHttp module

xavierb-amadeus:
JClouds is apparently exclusively using the Payload object from the HTTP
response to fill in the size of the BlobMetadata (when calling
blobStore.blobMetadata(...) ) - adapt this driver accordingly otherwise
we systematically get null size BlobMetadata out of it.


xavierb-amadeus:
this is a followup of this earlier pull request https://github.com/apache/jclouds/pull/48, unfortunately it was incomplete.

gaul:
Pushed to 2.2.x as 94f02856b7a9c891e63c2e7296d6ce00c70a43a3.  Thank you for your contribution @xavierb-amadeus!

PR #61, opened 2020-02-17
JCLOUDS-1533 - Fix upload with SAS token when blob name contains slash

majaseremet:


nacx:
Thanks!

nacx:
Cherry-picked to 2.2.x.

PR #64, opened 2020-03-09
JCLOUDS-1520 - Reuse SSLSocketFactory in JavaUrlHttpCommandExecutorSe…

roded:
…rvice

### DIFF
--- a/core/src/main/java/org/jclouds/http/config/SSLModule.java
+++ b/core/src/main/java/org/jclouds/http/config/SSLModule.java
@@ -16,9 +16,13 @@
  */
 package org.jclouds.http.config;
 
-import java.security.SecureRandom;
-import java.security.cert.X509Certificate;
-import java.util.Map;
+import com.google.common.base.Supplier;
+import com.google.common.base.Throwables;
+import com.google.common.collect.MapMaker;
+import com.google.inject.AbstractModule;
+import com.google.inject.TypeLiteral;
+import com.google.inject.name.Names;
+import org.jclouds.logging.Logger;
 
 import javax.annotation.Resource;
 import javax.inject.Inject;
@@ -26,25 +30,20 @@ import javax.inject.Singleton;
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLSession;
+import javax.net.ssl.SSLSocketFactory;
 import javax.net.ssl.TrustManager;
 import javax.net.ssl.X509TrustManager;
-
-import org.jclouds.logging.Logger;
-
-import com.google.common.collect.MapMaker;
-import com.google.common.base.Supplier;
-import com.google.common.base.Throwables;
-import com.google.inject.AbstractModule;
-import com.google.inject.TypeLiteral;
-import com.google.inject.name.Names;
+import java.security.SecureRandom;
+import java.security.cert.X509Certificate;
+import java.util.Map;
 
 public class SSLModule extends AbstractModule {
 
    @Override
    protected void configure() {
       bind(HostnameVerifier.class).annotatedWith(Names.named("untrusted")).to(LogToMapHostnameVerifier.class);
-      bind(new TypeLiteral<Supplier<SSLContext>>() {
-      }).annotatedWith(Names.named("untrusted")).to(new TypeLiteral<UntrustedSSLContextSupplier>() {
+      bind(new TypeLiteral<Supplier<SSLSocketFactory>>() {
+      }).annotatedWith(Names.named("untrusted")).to(new TypeLiteral<UntrustedSSLSocketFactorySupplier>() {
       });
    }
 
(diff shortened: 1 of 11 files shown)


roded:
Comments welcome. Thanks.

core/src/main/java/org/jclouds/http/internal/JavaUrlHttpCommandExecutorService.java
```diff
@@ -72,6 +74,7 @@
    protected Supplier<SSLContext> sslContextSupplier;
    protected final int outputSocketBufferSize;
    protected final String userAgent;
+   protected final AtomicReference<SSLSocketFactory> sslSocketFactoryReference;
```
nacx: I'm a bit unclear on this state on a singleton object. Could we just push back the decision on the SSL socket factory to use to the existing suppliers? If the suppliers are returning a singleton instance, as it is the case of the untrusted supplier, then we should be good here?
Instead of introducing state here, I'd just delegate that to the suppliers, as that is something that can be tweaked by users bu configuring their own ones when creating the context, if needed.

core/src/main/java/org/jclouds/http/internal/JavaUrlHttpCommandExecutorService.java
```diff
@@ -202,19 +206,38 @@ protected HttpURLConnection initConnection(HttpRequest request) throws IOExcepti
...
+      if (sslSocketFactory != null) {
+         sslCon.setSSLSocketFactory(sslSocketFactory);
+      }
+   }
+
+   private void setSslSocketFactoryReferenceIfAbsent(AtomicReference<SSLSocketFactory> sslSocketFactoryReference,
+                                                     Supplier<SSLContext> sslContextSupplier) {
+      if (sslSocketFactoryReference.get() == null) {
+         sslSocketFactoryReference.set(sslContextSupplier.get().getSocketFactory());
+      }
```
nacx: Apart from the comment above, this is racy.
```suggestion
      sslSocketFactoryReference.compareAndSet(null, sslContextSupplier.get().getSocketFactory());
```

core/src/main/java/org/jclouds/http/internal/JavaUrlHttpCommandExecutorService.java
roded: I don't think I can use the existing suppliers to return a singleton socket factory, SSLContext is not an interface I can easily implement. I could change all Supplier<SSLContext>s to Supplier<SSLSocketFactory>s, that way the supplier implementation could contain the state.

nacx: My main concern here is that the `JavaUrlHttpCommandExecutorService` is a singleton, and setting a member variable once will set its value for the entire lifecycle of the jclouds context. That means, every HTTP connection to any API method. Is this the intended behavior?

roded: Yeah that is intended - similarly to the way HttpsURLConnection uses the same instance of the default SSLSocketFactory (provided by SSLSocketFactory.getDefault()) in the case in which its SSLSocketFactory has not been explicitly set. Using the same SSLSocketFactory instance will allow HttpsClient to reuse existing connections from its KeepAliveCache.

I think that the better solution would be to inject Supplier<SSLSocketFactory>.

core/src/main/java/org/jclouds/http/internal/JavaUrlHttpCommandExecutorService.java
roded: No longer relevant.

roded:
I've pushed an update as per the discussion above. It's still not tested though - I'll update once it is.

roded:
Tested to be working on OpenJDK 8 and OpenJ9 8.
Affected modules' test suites pass OK.

core/src/main/java/org/jclouds/http/config/SSLModule.java
```diff
@@ -43,8 +44,8 @@
    @Override
    protected void configure() {
       bind(HostnameVerifier.class).annotatedWith(Names.named("untrusted")).to(LogToMapHostnameVerifier.class);
-      bind(new TypeLiteral<Supplier<SSLContext>>() {
-      }).annotatedWith(Names.named("untrusted")).to(new TypeLiteral<UntrustedSSLContextSupplier>() {
+      bind(new TypeLiteral<Supplier<SSLSocketFactory>>() {
+      }).annotatedWith(Names.named("untrusted")).to(new TypeLiteral<UntrustedSSLSocketFactorySupplier>() {
```
nacx, line 48: I'm a bit concerned about this change. There are still uses of the SSLContext injection. For example, take a look at the AapcheHC or the OkHttp drivers. Those would need to be updated too.

I can also imagine that there can also be users that are providing a custom version of the SSLContext, to customize how it behaves, and I think we should try not to break them. Is there any way you can introduce your changes by keeping the old supplier? Would it be possible to have both, the old one and your SSLSocketFactory new one, and let your supplier use the injected SSLContext one? This way we keep backward-compatibility and the only thing we care about is our codebase using the factory supplier instead of directly getting it from the context.

roded: I understand your concern. I could have JavaUrlHttpCommandExecutorService accept both a nullable Supplier<SSLContext> and a Supplier<SSLSocketFactory> for the untrusted use case as a fallback - this will be in line with the existing behavior and shouldn't break existing users providing SSLContext. In this case though, I suspect that only the untrusted use case will be using the socket cache.

I'm not sure if the untrusted SSLSocketFactory supplier should build upon another injected SSLContext supplier since these are currently mutually exclusive use cases in JavaUrlHttpCommandExecutorService (and also I don't think SSLContext can be init-ed twice).

roded:
I've pushed another attempt at this which handles the issue only for the untrusted use case. 
JavaUrlHttpCommandExecutorService now keeps its existing optional (trusted) Supplier<SSLContext> and fallbacks to a Supplier<SSLSocketFactory> for the untrusted usecase.

Classes extending JavaUrlHttpCommandExecutorService have been changed to support the new constructor signature. 

LazyCachingSSLSocketFactorySupplier was introduced to keep the SSLSocketFactory creation during first client access instead of during the Guice Module's initialization.

I'm currently having some trouble building the project locally due to missing gson packages (?). I'll have to see how to work around that.

nacx:
Hi! Apologies for the late review. I've been really busy and had no cycles to look at this PR again.

The fix looks good. So the summary is that we need to provide a singleton SSLContext, but also a singleton SSLContext factory. The current approach, however, just addresses this for the `untrusted` context. Would the issue also manifest when a proper, trusted SSLContext is provided? If so, we should have feature parity there, and allow also to set the custom SSLSocketFactory supplier. One option could be to have the Lazy supplier you created implement both suppliers, and allow users to easily use it to inject both consistently where needed?

gaul:
@roded We plan to release 2.3.0 later this month.  Do you think you can finish this PR soon?

roded:
@gaul I won't be able to complete this PR for 2.3.0.
Would you have me close this PR and keep the issue open?

Sorry for leaving this open for so long.

gaul:
@roded Do we have a path forward on this PR?

roded:
@gaul Due to time constraints, I don't have a path forward for this PR unfortunately. Apologies.

gaul:
Closing due to inactivity.

PR #65, opened 2020-04-03
JCLOUDS-1541: Add Middle East (Bahrain) region to the AWS EC2 and S3 providers list

ikky888:
This change request is to add the Middle East (Bahrain) region to the AWS EC2 and S3 providers list.

I was able to compile it on my forked repo but as this is my first pull request, I want to make sure I missed nothing.


gaul:
Also pushed to 2.2.x as f8768003708d78fcb599373523599f2bed63cf27.  Thank you for your contribution @ikky888!

PR #66, opened 2020-04-14
JCLOUDS-1543: change FetchBlobMetadata to retain original blob order

roded:


gaul:
Also pushed to master as 5ac92111c4e9347c492a16ecd402acafdb1f9cd7.  Thank you for your contribution @roded!

PR #67, opened 2020-04-16
Fix javadoc generation on JDK >= 8

felfert:
This fixes javadoc generation with JDK >= 8 so that someone with write-permission to the jclouds.apache.org site can fix apache/jclouds-site#12

Corresponding JIRA-Issue: [JCLOUDS-1544](https://issues.apache.org/jira/browse/JCLOUDS-1544)
Note: This also applies to the 2.2.x branch, shall I create another PR for that?

Cheers
 -Fritz

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -83,6 +83,9 @@
         <plugins>
           <plugin>
             <artifactId>maven-javadoc-plugin</artifactId>
+            <configuration>
+              <additionalparam>${javadoc.opts}</additionalparam>
+            </configuration>
             <executions>
               <execution>
                 <id>javadoc</id>
--- a/project/pom.xml
+++ b/project/pom.xml
@@ -886,6 +886,7 @@
                   <encoding>${project.build.sourceEncoding}</encoding>
                   <quiet>true</quiet>
                   <maxmemory>512m</maxmemory>
+                  <additionalparam>${javadoc.opts}</additionalparam>
                 </configuration>
                 <reportSets>
                   <reportSet>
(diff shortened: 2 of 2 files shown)


felfert:
Another note:
Obviously, this currently does not get validated in travis. I could add a modification to travis.yml, but I'm not shure if that is desired. IMHO, it is better to create another PR **after** this one gets merged.

nacx:
>Obviously, this currently does not get validated in travis. I could add a modification to travis.yml, but I'm not shure if that is desired. IMHO, it is better to create another PR after this one gets merged.

+1, that looks good

gaul:
Pushed to 2.2.x as 876269427b80d1d4a68403e0e729342d340fbd82.

gaul:
@felfert @nacx Oops I missed that this broke master:

https://builds.apache.org/job/jclouds/101/console

It appears that we build with Java 7 and need some profile selection to fix it.  @felfert can you look into this?

felfert:
Thats strange. I **did** create a profile "disable-doclint" which should get activated by jdk8 only.
Even stranger, in travis.yml it says:
```
jdk:
  - openjdk8
```

Well, I guess I have to fetch an old jdk to verify this...

felfert:
Well, on a clean VM with JDK7 (java-1.7.0-openjdk-devel-1.7.0.251-2.6.21.0.el7_7.x86_64)
everything builds fine here. Since I don't have access to your jenkins setup, I can only guess, that this setup might be broken in that respect. In fact, the build log shows, that jdk**1.8** is used to run maven but then this is apparently reverted by  using -Pjenkins which activates maven-toolchains-plugin which itself uses **~/.m2/toolchains.xml** to which I don't have access. So my hands are tied here.

=> Can one of you guys can provide me with that file so I can investigate further?
@nacx : Maybe you introduced this jenkins profile for exactly that reason (Use jdk7 in order to solve the javadoc errors with JDK8 on jenkins)? If so, you could simply drop the -Pjenkins argument?

felfert:
@nacx , @gaul 
See #68

PR #71, opened 2020-04-30
JCLOUDS-1545: Upgrade to OkHttp 2.7.5

gaul:
This allows creating an SSL certificate with 2048-bit RSA which
enables newer Java versions to run integration tests.  Tweak okhttp
tests which do not allow zero-length requests without bodies.


nacx:
Hmmm... I've never seen that one, but it looks indeed like a version incompatibility with BC?

gaul:
I looked into this a little further.  OkHttp uses bcprov-jdk15on 1.65 but aligning our version doesn't resolve the symptoms.

gaul:
Never mind -- I was using an older version of OkHttp, 2.7.5.  Let me work forward...

gaul:
Superseded by #84.

PR #73, opened 2020-05-23
JCLOUDS-912: JCLOUDS-1547: GCS InputStream single-part upload

gaul:
Previously this provider worked around a RestAnnotationProcessor quirk
by using multi-part uploads for InputStream payloads.  Instead work
around the quirk another way which allows a single-part upload.  This
allows inclusion of the Content-MD5 header during object creation.
Backfill tests with both ByteSource and InputStream inputs.


providers/google-cloud-storage/src/main/java/org/jclouds/googlecloudstorage/binders/MultipartUploadBinder.java
```diff
@@ -57,6 +57,7 @@
       Part jsonPart = Part.create("Metadata", jsonPayload, new Part.PartOptions().contentType(APPLICATION_JSON));
       Part mediaPart = Part.create(template.name(), payload, new Part.PartOptions().contentType(contentType));
 
+      request.resetPayload(/*release=*/ false);
```
gaul, line 60: This is not ideal but I don't understand what the `RestAnnotationProcessor` is doing.

nacx: Sorry I'm a bit late here. Do you have a set of specific lines/methods on the processor? Perhaps I can help fix those instead.

PR #75, opened 2020-06-14
JCLOUDS-1333: JCLOUDS-1334: JCLOUDS-1470: Require Java 8 and Guava 22

gaul:
This allows compatibility with Guava 29.  Also unwind some older
workarounds.

### DIFF
--- a/project/pom.xml
+++ b/project/pom.xml
@@ -217,13 +217,13 @@
   </developers>
 
   <properties>
-    <maven.compile.source>1.7</maven.compile.source>
-    <maven.compile.target>1.7</maven.compile.target>
-    <jdk.version>1.7</jdk.version>
+    <maven.compile.source>1.8</maven.compile.source>
+    <maven.compile.target>1.8</maven.compile.target>
+    <jdk.version>1.8</jdk.version>
     <maven.compile.deprecation>true</maven.compile.deprecation>
     <maven.site.url.base>gitsite:git@github.com/jclouds/jclouds-maven-site.git</maven.site.url.base>
-    <guava.version>18.0</guava.version>
-    <guava.osgi.import>com.google.common.*;version="[18.0,24.0.0)"</guava.osgi.import>
+    <guava.version>22.0</guava.version>
+    <guava.osgi.import>com.google.common.*;version="[22.0,30.0.0)"</guava.osgi.import>
     <guice.version>3.0</guice.version>
     <okhttp.version>2.2.0</okhttp.version>
     <okio.osgi.import>okio;version="[1.2.0,1.3)"</okio.osgi.import>
@@ -585,7 +585,7 @@
             <configuration>
               <rules>
                 <requireJavaVersion>
-                  <version>[1.7,)</version>
+                  <version>[1.8,)</version>
                 </requireJavaVersion>
                 <requireMavenVersion>
                   <version>[3.0.2,)</version>
--- a/allblobstore/pom.xml
+++ b/allblobstore/pom.xml
@@ -74,19 +74,4 @@
       <version>${project.version}</version>
     </dependency>
   </dependencies>
-  <profiles>
-    <profile>
-      <id>jre1.7</id>
-      <activation>
-        <jdk>[1.7,)</jdk>
-      </activation>
-      <dependencies>
-        <dependency>
-          <groupId>org.apache.jclouds.api</groupId>
-          <artifactId>filesystem</artifactId>
-         <version>${project.version}</version>
-       </dependency>
-      </dependencies>
-    </profile>
-  </profiles>
 </project>
--- a/apis/chef/src/test/java/org/jclouds/chef/features/OrganizationApiExpectTest.java
+++ b/apis/chef/src/test/java/org/jclouds/chef/features/OrganizationApiExpectTest.java
@@ -34,6 +34,7 @@ import org.jclouds.chef.domain.User;
 import org.jclouds.date.TimeStamp;
 import org.jclouds.http.HttpRequest;
 import org.jclouds.http.HttpResponse;
+import org.jclouds.http.HttpResponseException;
 import org.jclouds.rest.ConfiguresHttpApi;
 import org.jclouds.rest.ResourceNotFoundException;
 import org.testng.annotations.Test;
@@ -144,7 +145,7 @@ public class OrganizationApiExpectTest extends BaseChefApiExpectTest<ChefApi> {
       api.organizationApi().get().deleteGroup("foo");
    }
 
-   @Test(expectedExceptions = ResourceNotFoundException.class)
+   @Test(expectedExceptions = HttpResponseException.class)
    public void testDeleteGroupFailsOn404() {
       ChefApi api = requestSendsResponse(signed(HttpRequest.builder() //
             .method("DELETE") //
--- a/apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/ImageToOperatingSystem.java
+++ b/apis/openstack-nova/src/main/java/org/jclouds/openstack/nova/v2_0/compute/functions/ImageToOperatingSystem.java
@@ -85,7 +85,7 @@ public class ImageToOperatingSystem implements Function<Image, OperatingSystem> 
          } else if (imageName.contains("Oracle")) {
             osFamily = OsFamily.OEL;
          } else {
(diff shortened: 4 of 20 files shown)


gaul:
This will likely fail in CI due to configuration but locally tests fail due a gson issue.  I don't understand our tower of hacks.  @nacx Any suggestions on how to address this?

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.17:test (default-test) on project jclouds-core: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.17:test failed: There was an error in the forked process
[ERROR] java.lang.NoClassDefFoundError: com/google/gson/TypeAdapterFactory
[ERROR]         at java.lang.ClassLoader.defineClass1(Native Method)
[ERROR]         at java.lang.ClassLoader.defineClass(ClassLoader.java:756)
[ERROR]         at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
...
[ERROR] Caused by: java.lang.ClassNotFoundException: com.google.gson.TypeAdapterFactory
[ERROR]         at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
[ERROR]         at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
[ERROR]         at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352)
[ERROR]         at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
[ERROR]         ... 39 more
```

gaul:
This is strange, `mvn test` cannot find gson but `mvn install` makes progress until failing tests.

gaul:
Also need to use an older JDK to work around weak 1024-bit SSL certificates.  I found that `OrganizationApiExpectTest.testDeleteGroupFailsOn404` fails due to special handling in `ChefErrorHandler.handleError`.  Commenting this out allows the test to succeed but I don't understand how it passed before.

apis/chef/src/main/java/org/jclouds/chef/handlers/ChefErrorHandler.java
```diff
@@ -57,9 +57,7 @@ public void handleError(HttpCommand command, HttpResponse response) {
                exception = new AuthorizationException(message, exception);
                break;
             case 404:
-               if (!command.getCurrentRequest().getMethod().equals("DELETE")) {
-                  exception = new ResourceNotFoundException(message, exception);
-               }
+               exception = new ResourceNotFoundException(message, exception);
```
gaul, line 60: @nacx Any comment on this change?

project/pom.xml
```diff
@@ -217,12 +217,12 @@
...
-    <maven.compile.source>1.7</maven.compile.source>
-    <maven.compile.target>1.7</maven.compile.target>
-    <jdk.version>1.7</jdk.version>
+    <maven.compile.source>1.8</maven.compile.source>
+    <maven.compile.target>1.8</maven.compile.target>
+    <jdk.version>1.8</jdk.version>
     <maven.compile.deprecation>true</maven.compile.deprecation>
     <maven.site.url.base>gitsite:git@github.com/jclouds/jclouds-maven-site.git</maven.site.url.base>
-    <guava.version>18.0</guava.version>
+    <guava.version>22.0</guava.version>
```
gaul, line 225: We could support 21.0 with some workarounds for `SimpleTimeLimiter` although this doesn't seem worth the effort.

nacx:
There were some [gson hacks](https://github.com/apache/jclouds/tree/master/gson) we did to be able to make it OSGi friendly, as there are uses of an `internal` gson package we use that has been unexpected in newer versions of the library. That made it impossible to upgrade to newer versions of it.

Basically, we shaded those packages and renamed them so we can safely use them in our codebase.

Now we're in a different place, and I think we should revisit our OSGi support. We moved jclouds-Karaf to the Karaf project because we lack the expertise in OSGi and maintaining it was adding too much burden on us. Does the same apply here? Should we just undo this hack, go with standard gson and upgrade normally, and remove the OSGi compatibility from our list of responsibilities? This shading hack is something that can be done by users, or even by the Karaf project, probably in a better way, so I'd be in favour of undoing this at the expense of losing compatibility. The cost of maintaining this is too high for the current community.

apis/chef/src/main/java/org/jclouds/chef/handlers/ChefErrorHandler.java
nacx: Oh, I see that it is because of a test. I'd say the test is wrong, not this, so I'd just change the test and keep the current behavior.

nacx:
>Also need to use an older JDK to work around weak 1024-bit SSL certificates.

Is this because of the certificates we use in tests? Or something spotted running live tests, or because how we configure the SSL module by default? It would be great if we could upgrade and not be tied to older JDKs.

gurkerl83:
Hello, we use your JClouds in our OSGi project. Actually OSGi and all the regulations around packet exports is a difficult topic. Especially when an OSGi compatibility is set by a library and others still access restricted areas of the library. 

I can't understand why the makers of JSON restrict these sections, instead every package of the library should be exported, after all there is not much to hide 

- internal/ 
- internal/bind
- internal/bind/utile 
- internal/reflect

I'm not sure if GSON really intends to keep these packages away from the outside world, or if this is just a misconfiguration. You can try submitting a merge request to GSON by exporting all packages, adding four lines to the "exportcontents" section of the bnd.bnd configuration file Due to other experiences with JSON I can't promise if a change of this kind will be accepted by the developers. After all, very few Java developers know about OSGi and the rules of the construct, developers working at Google are no exception.

The proposal to let the user override the GSON module is basically good. The way of proceeding is left to him, but it should be comprehensible, the requirement to do so should be documented in JClouds in any case. My approach at this point is to create a new git project, integrate a git submodule into GSON, and use a script (changing the BND file before build) or a Maven plugin to influence the build process of GSON. An example of where we do this is https://github.com/project-millipede/neo4j-ogm-osgi. Since GSON releases are not released too often, updating the submodule is not a hassle.

I am interested in the process, whatever this will look like, I am curious.

nacx:
>I'm not sure if GSON really intends to keep these packages away from the outside world, or if this is just a misconfiguration. You can try submitting a merge request to GSON by exporting all packages, adding four lines to the "exportcontents" section of the bnd.bnd configuration file Due to other experiences with JSON I can't promise if a change of this kind will be accepted by the developers. After all, very few Java developers know about OSGi and the rules of the construct, developers working at Google are no exception.

We tried, but unfortunately, it was rejected.

The main issue here is that we can no longer afford to keep the OSGi compatibility because of the lack of expertise in the active contributor group to the project. That's why we offloaded the Karaf support to the Karaf project, and this is a remaining piece that is preventing us from modernizing the codebase. Help there is really welcome.

Ideally, we should try to refactor the jclouds core to stop relying on that, or even copy the files to the jclouds codebase and keep them there, but I think we can't afford the overhead of keeping OSGi into account when the expertise is not here.

apis/chef/src/main/java/org/jclouds/chef/handlers/ChefErrorHandler.java
gaul: Done.

gaul:
> Is this because of the certificates we use in tests? Or something spotted running live tests, or because how we configure the SSL module by default? It would be great if we could upgrade and not be tied to older JDKs.

#71 should resolve the certificate length but we need to upgrade BouncyCastle first.   I don't remember exactly but that was blocked on another dependency.  Working through these...

gaul:
> Hello, we use your JClouds in our OSGi project. Actually OSGi and all the regulations around packet exports is a difficult topic. Especially when an OSGi compatibility is set by a library and others still access restricted areas of the library.

@gurkerl83 thanks for chiming in.  We lack OSGi expertise and would appreciate if you could submit any improvements.  I am trying to modernize the dependencies and fear that this will break things I cannot test like OSGi.

gurkerl83:
Hi, I have create a follow up JIRA ticket to summarise the GSON problematic https://issues.apache.org/jira/browse/JCLOUDS-1548. Further I wanted to know when the current merge request will be merged. I have also reviewed the changes, which in my opinion are legit, also all tests pass.
Thx!

gaul:
@nacx Any further comments?

nacx:
Nope, LGTM. Thanks!

cdancy:
@gaul @nacx any chance we can get a new release with this fix?

gaul:
Probably a few months.  We are still working on the OkHttp and gson dependencies.

PR #76, opened 2020-06-23
JCLOUDS-1542 Java 11 warns of illegal reflective access

JnRouvignac:
With Java 11, an illegal reflective access is output for the google cloud storage blobstore.

    WARNING: An illegal reflective access operation has occurred
    WARNING: Illegal reflective access by org.jclouds.reflect.Reflection2$1 (file:/.../jclouds-core.jar) to constructor java.lang.String(char[],int,int,java.lang.Void)
    WARNING: Please consider reporting this to the maintainers of org.jclouds.reflect.Reflection2$1
    WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
    WARNING: All illegal access operations will be denied in a future release

Indeed, JClouds calls `setAccessible(true)` on the package protected constructor
     `java.lang.String(char[],int,int,java.lang.Void)`.
Investigating the code, it turns out it is looking for constructors annotated with any of:
* java.beans.ConstructorProperties
* org.jclouds.json.SerializedNames
* com.google.inject.Inject

But `String` being defined in `java.base` module, it is impossible
     that it will be annotated with any of these annotation.

This commit is complementary to JClouds commit db4e4af931ef19f582b85f02efb93e50f1c5d01c .

&nbsp;

Reflection2.java:
Do not call `setAccessible(true)` on core java constructors and methods.

&nbsp;

For reference, here is the stacktrace of this illegal access warning:

    java.lang.String.<init>(String.java:3208)
    java.lang.String.<init>(String.java:251)
    java.util.StringJoiner.compactElts(StringJoiner.java:250)
    java.util.StringJoiner.toString(StringJoiner.java:173)
    jdk.internal.module.IllegalAccessLogger.loudWarning(IllegalAccessLogger.java:339)
    jdk.internal.module.IllegalAccessLogger.log(IllegalAccessLogger.java:288)
    jdk.internal.module.IllegalAccessLogger.log(IllegalAccessLogger.java:261)
    jdk.internal.module.IllegalAccessLogger.logIfOpenedForIllegalAccess(IllegalAccessLogger.java:226)
    java.lang.reflect.AccessibleObject.logIfOpenedForIllegalAccess(AccessibleObject.java:366)
    java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:325)
    java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:280)
    java.lang.reflect.Constructor.checkCanSetAccessible(Constructor.java:189)
    java.lang.reflect.Constructor.setAccessible(Constructor.java:182)
    org.jclouds.reflect.Reflection2$1.load(Reflection2$1.java:157)
    org.jclouds.reflect.Reflection2$1.load(Reflection2$1.java:153)
    com.google.common.cache.LocalCache$LoadingValueReference.loadFuture(LocalCache$LoadingValueReference.java:3529)
    com.google.common.cache.LocalCache$Segment.loadSync(LocalCache$Segment.java:2278)
    com.google.common.cache.LocalCache$Segment.lockedGetOrLoad(LocalCache$Segment.java:2155)
    com.google.common.cache.LocalCache$Segment.get(LocalCache$Segment.java:2045)
    com.google.common.cache.LocalCache.get(LocalCache.java:3953)
    com.google.common.cache.LocalCache.getOrLoad(LocalCache.java:3976)
    com.google.common.cache.LocalCache$LocalLoadingCache.get(LocalCache$LocalLoadingCache.java:4960)
    org.jclouds.reflect.Reflection2.get(Reflection2.java:346)
    org.jclouds.reflect.Reflection2.constructors(Reflection2.java:100)
    org.jclouds.json.internal.NamingStrategies$AnnotationConstructorNamingStrategy.getDeserializer(NamingStrategies$AnnotationConstructorNamingStrategy.java:271)
    org.jclouds.json.internal.DeserializationConstructorAndReflectiveTypeAdapterFactory.create(DeserializationConstructorAndReflectiveTypeAdapterFactory.java:125)
    com.google.gson.Gson.getAdapter(Gson.java:458)
    org.jclouds.json.gson.internal.bind.ReflectiveTypeAdapterFactory.createBoundField(ReflectiveTypeAdapterFactory.java:117)
    org.jclouds.json.gson.internal.bind.ReflectiveTypeAdapterFactory.getBoundFields(ReflectiveTypeAdapterFactory.java:166)
    org.jclouds.json.gson.internal.bind.ReflectiveTypeAdapterFactory.create(ReflectiveTypeAdapterFactory.java:102)
    com.google.gson.Gson.getAdapter(Gson.java:458)
    com.google.gson.Gson.toJson(Gson.java:696)
    com.google.gson.Gson.toJson(Gson.java:683)
    com.google.gson.Gson.toJson(Gson.java:638)
    com.google.gson.Gson.toJson(Gson.java:618)
    org.jclouds.json.internal.GsonWrapper.toJson(GsonWrapper.java:65)
    org.jclouds.oauth.v2.functions.ClaimsToAssertion.apply(ClaimsToAssertion.java:59)
    org.jclouds.oauth.v2.functions.ClaimsToAssertion.apply(ClaimsToAssertion.java:43)
    org.jclouds.rest.internal.RestAnnotationProcessor.getParamValue(RestAnnotationProcessor.java:829)
    org.jclouds.rest.internal.RestAnnotationProcessor.getFormParamKeyValues(RestAnnotationProcessor.java:847)
    org.jclouds.rest.internal.RestAnnotationProcessor.addFormParams(RestAnnotationProcessor.java:435)
    org.jclouds.rest.internal.RestAnnotationProcessor.apply(RestAnnotationProcessor.java:258)
    org.jclouds.rest.internal.RestAnnotationProcessor.apply(RestAnnotationProcessor.java:137)
    org.jclouds.rest.internal.InvokeHttpMethod.toCommand(InvokeHttpMethod.java:189)
    org.jclouds.rest.internal.InvokeHttpMethod.invoke(InvokeHttpMethod.java:85)
    org.jclouds.rest.internal.InvokeHttpMethod.apply(InvokeHttpMethod.java:74)
    org.jclouds.rest.internal.InvokeHttpMethod.apply(InvokeHttpMethod.java:45)
    org.jclouds.rest.internal.DelegatesToInvocationFunction.handle(DelegatesToInvocationFunction.java:156)
    org.jclouds.rest.internal.DelegatesToInvocationFunction.invoke(DelegatesToInvocationFunction.java:123)
    com.sun.proxy.$Proxy49.authorize(Unknown Source)
    org.jclouds.oauth.v2.filters.JWTBearerTokenFlow$AuthorizeToken.load(JWTBearerTokenFlow$AuthorizeToken.java:84)
    org.jclouds.oauth.v2.filters.JWTBearerTokenFlow$AuthorizeToken.load(JWTBearerTokenFlow$AuthorizeToken.java:68)
    com.google.common.cache.LocalCache$LoadingValueReference.loadFuture(LocalCache$LoadingValueReference.java:3529)
    com.google.common.cache.LocalCache$Segment.loadSync(LocalCache$Segment.java:2278)
    com.google.common.cache.LocalCache$Segment.lockedGetOrLoad(LocalCache$Segment.java:2155)
    com.google.common.cache.LocalCache$Segment.get(LocalCache$Segment.java:2045)
    com.google.common.cache.LocalCache.get(LocalCache.java:3953)
    com.google.common.cache.LocalCache.getOrLoad(LocalCache.java:3976)
    com.google.common.cache.LocalCache$LocalLoadingCache.get(LocalCache$LocalLoadingCache.java:4960)
    com.google.common.cache.LocalCache$LocalLoadingCache.getUnchecked(LocalCache$LocalLoadingCache.java:4966)
    org.jclouds.oauth.v2.filters.JWTBearerTokenFlow.filter(JWTBearerTokenFlow.java:99)
    org.jclouds.http.internal.BaseHttpCommandExecutorService.invoke(BaseHttpCommandExecutorService.java:90)
    org.jclouds.rest.internal.InvokeHttpMethod.invoke(InvokeHttpMethod.java:91)
    org.jclouds.rest.internal.InvokeHttpMethod.apply(InvokeHttpMethod.java:74)
    org.jclouds.rest.internal.InvokeHttpMethod.apply(InvokeHttpMethod.java:45)
    org.jclouds.reflect.FunctionalReflection$FunctionalInvocationHandler.handleInvocation(FunctionalReflection.java:117)
    com.google.common.reflect.AbstractInvocationHandler.invoke(AbstractInvocationHandler.java:86)
    com.sun.proxy.$Proxy54.simpleUpload(Unknown Source)
    org.jclouds.googlecloudstorage.blobstore.GoogleCloudStorageBlobStore.uploadMultipartPart(GoogleCloudStorageBlobStore.java:422)
    org.jclouds.blobstore.internal.BaseBlobStore$BlobUploader.call(BaseBlobStore.java:415)
    org.jclouds.blobstore.internal.BaseBlobStore$BlobUploader.call(BaseBlobStore.java:402)
    com.google.common.util.concurrent.TrustedListenableFutureTask$TrustedFutureInterruptibleTask.runInterruptibly(TrustedListenableFutureTask.java:125)
    com.google.common.util.concurrent.InterruptibleTask.run(InterruptibleTask.java:69)
    com.google.common.util.concurrent.TrustedListenableFutureTask.run(TrustedListenableFutureTask.java:78)
    java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
    java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor$Worker.java:628)
    java.lang.Thread.run(Thread.java:834)

### DIFF
--- a/core/src/main/java/org/jclouds/reflect/Reflection2.java
+++ b/core/src/main/java/org/jclouds/reflect/Reflection2.java
@@ -153,15 +153,23 @@ public class Reflection2 {
          .newBuilder().build(new CacheLoader<TypeToken<?>, Set<Invokable<?, ?>>>() {
             public Set<Invokable<?, ?>> load(TypeToken<?> key) {
                ImmutableSet.Builder<Invokable<?, ?>> builder = ImmutableSet.<Invokable<?, ?>> builder();
-               for (Constructor<?> ctor : key.getRawType().getDeclaredConstructors()) {
-                  ctor.setAccessible(true);
+               Class<?> raw = key.getRawType();
+               for (Constructor<?> ctor : raw.getDeclaredConstructors()) {
+                  if (!coreJavaClass(raw)) {
+                     // In JDK 11 up to 14, the only uses for `java.beans.ConstructorProperties` annotation
+                     // are in the `java.awt`, `java.beans` and `javax.swing` packages.
+                     // Since these constructors are public, there is no need to call `setAccessible()`
+                     ctor.setAccessible(true);
+                  }
                   builder.add(key.constructor(ctor));
                }
                // Look for factory methods, if this is an abstract type.
-               if (Modifier.isAbstract(key.getRawType().getModifiers())) {
-                  for (Invokable<?, Object> method : methods(key.getRawType())){
+               if (Modifier.isAbstract(raw.getModifiers())) {
+                  for (Invokable<?, Object> method : methods(raw)){
                      if (method.isStatic() && method.getReturnType().equals(key)) {
-                        method.setAccessible(true);
+                        if (!coreJavaClass(raw)) {
+                           method.setAccessible(true);
+                        }
                         builder.add(method);
                      }
                   }


core/src/main/java/org/jclouds/reflect/Reflection2.java
```diff
@@ -153,15 +153,23 @@
...
             public Set<Invokable<?, ?>> load(TypeToken<?> key) {
                ImmutableSet.Builder<Invokable<?, ?>> builder = ImmutableSet.<Invokable<?, ?>> builder();
-               for (Constructor<?> ctor : key.getRawType().getDeclaredConstructors()) {
-                  ctor.setAccessible(true);
+               Class<?> raw = key.getRawType();
+               for (Constructor<?> ctor : raw.getDeclaredConstructors()) {
+                  if (!coreJavaClass(raw)) {
+                     // In JDK 11 up to 14, the only uses for `java.beans.ConstructorProperties` annotation
+                     // are in the `java.awt`, `java.beans` and `javax.swing` packages.
+                     // Since these constructors are public, there is no need to call `setAccessible()`
```
nacx, line 161: Instead of doing this, could we just do `if (!ctor.isAccessible()) ctor.setAccessible(true)`? Probably this is better for the general case and makes the code safer (and this comment irrelevant)?
We can keep the core java class checks if needed, but it looks a bit safer to just check if it is accessible, then try setting it. And if not accessible and a core Java class, we can probably log a warning message as something is probably gonna fail later as we can't set the proper access level tot he method?

JnRouvignac: Given the javadoc of  [`AccessibleObject.isAccessible()`](https://docs.oracle.com/javase/9/docs/api/java/lang/reflect/AccessibleObject.html#isAccessible--), it looks almost certain that `ctor.isAccessible()` will return false for that package private constructor of String, and so `ctor.setAccessible(true)` will be called anyway and we will get the illegal access warning again. Our code is not calling  `ctor.setAccessible(true)` at all.

Even when using [`AccessibleObject.canAccess()`](https://docs.oracle.com/javase/9/docs/api/java/lang/reflect/AccessibleObject.html#canAccess-java.lang.Object-), it will return false in our case and code will end up calling `ctor.setAccessible(true)`, thus we will still get the illegal access warning.

Am I missing or misunderstanding something?

JnRouvignac: I read again the code and your comment, and I think you are saying I could drop the last sentence of the comment and then do the following?
```java
                  if (!ctor.canAccess(null) && !coreJavaClass(raw)) {
                     // In JDK 11 up to 14, the only uses for `java.beans.ConstructorProperties` annotation
                     // are in the `java.awt`, `java.beans` and `javax.swing` packages.
                     ctor.setAccessible(true);
                  }
```
We need to test it, but that should work indeed. There would be less spurious warnings for regular cases.

The check on `coreJavaClass()` is still required in our case though.

demobox: > Instead of doing this, could we just do if (!ctor.isAccessible()) ctor.setAccessible(true)?

@nacx That would indeed be cleaner, but I suspect (as I think @JnRouvignac is describing) this will not fix the warnings:

* jclouds explicitly tries to make certain non-accessible members accessible via reflection, and needs this to work
* **some** of those non-accessible members can be made accessible, **some** trigger warnings
* the latter set contains core classes that jclouds actually does **not** need to make accessible, so it should be safe to skip them

I no longer recall why we didn't add the "skip core classes" check to `constructorsForTypeToken` when we added it to `methodsForTypeToken` in db4e4af (I suspect it wasn't required to fix whatever issue we were seeing then?); it seems consistent enough to me to add it now, assuming all tests pass.

I agree that your suggestion - only try to make members accessible that aren't *already* accessible - would be a logic improvement, as we would only try to modify members where needed. I wonder, though, if it may actually be less performant, given that `isAccessible` would have to be invoked many times?

I think it makes sense to consider that potential improvement - here and also in `methodsForTypeToken` - as a separate PR if desired, though.

Happy to discuss further if useful!

JnRouvignac: Note 1: `isAccessible()` is not useful. I think you meant `canAccess()`.
Note 2: if if understand correctly, this code does not look performance sensitive: it is only called once for each `TypeToken` (i.e. for each class?) in the lifetime of an application, since I see no expiration for this cache.

> I think it makes sense to consider that potential improvement - here and also in methodsForTypeToken - as a separate PR if desired, though.

I would also prefer to have the original PR in as it fixes the problem, then I am also happy to follow-up with another to add the `canAccess()` checks everywhere.
Let me know how you want to proceed.

nacx: Thanks for the replies, and yes, that makes sense. Let's keep the original proposed change as it addresses the issue and it aligns with the existing approach in the other method Andrew mentioned.

JnRouvignac:
No problem. Thank you for taking the time to review it!
I'll follow up with the another Pr with the `!canAccess()` when this one is merged.

nacx:
Thanks!

PR #78, opened 2020-06-25
Replace embedded and repackaged GSON library

gurkerl83:
Replace substituted GSON package names with those provided from the vendor.
Reduce OSGi-metadata declaration of core-module because the artificial package org.jclouds.json.gson.internal was removed.
Remove the Gson module its children Gson bundle, and Gson shaded.
Remove duplication conflict and check-style rules due to the removal of the internal Gson module.
Add maven repository where a custom version of the Gson library gets hosted, which exports all packages.


gurkerl83:
This is the first in a series of merge requests, which tries to mitigate the somewhat difficult behavior when dealing with OSGi incompatible libraries. Basically GSON is compatible with OSGi, but the way JClouds currently uses GSON reverses this characteristic.

Basically the OSGi problem only becomes apparent at runtime of an application. For this reason, the GSON library included in this merge request is the version of the manufacturer.

Nevertheless, a specially created repository was redefined in the section of the Maven declaration repositories. In this repository, the GSON library is located, with the corresponding changes to export all packages of the library. If desired, this can be removed again.

Read more about it here, including the auto release process.
https://github.com/project-millipede/gson-osgi

Feedback is very welcome

project/pom.xml
```diff
@@ -86,6 +86,16 @@
   </scm>
 
   <repositories>
+    <repository>
+      <id>gitlab-maven</id>
+      <url>https://gitlab.com/api/v4/projects/19575612/packages/maven</url>
```
gaul, line 91: @nacx will Apache let us depend on a package outside of Maven Central?  This seems dicey in any case.

nacx:
I think this is not correct. The project depends on pulling a dependency from a specific repository, otherwise, it will break.
The main issue here is releasing under the same `groupId` and `artifactId`, something that is not official. That is what Maven coordinates are for, so people can understand provenance and trust the libraries. We shouldnt' have a dependency called gson if it is not the official gown, nor we should depend on being connected to that particular unofficial repository.

The previous approach was renaming and repackaging, just to make sure it was very clear that we were using a modified gson bundle. If this becomes a problem, then I'd consider just reverting to official gson, and make this an OSGi consumers problem. One good step, would be to contribute the updated bundle to `Servicemix`, for example that has a lot of convenience dependencies for OSGi, and pull it from there, but in the end, whatever approach we pick, we shouldn't be _replacing_ official libraries in this way.

project/pom.xml
nacx: So, first, we need to make sure packages there are ASLv2 compliant, and then yes, we need to make sure we pull dependencies from a source that is guaranteed _to be there's forever. Apache releases must be always buildable from source, and I'm not sure if depending on this would meet these guarantees.
Apart from this, I've already commented on my main concerns in the PR.

gurkerl83: True, in the long run, and to comply with Apache's terms, specifying a particular repository is not the best choice. Let's use GSON in the original, which is distributed through the standard distribution channel. 

In any case, the consumer is responsible for whether the runtime environment requires a customized bundle or uses the one in the specified repository.

In the context of OSGi, e.g. when creating Karaf assemblies or features, a specific repository can always be specified from where a critical and customized library is loaded.

Translated with www.DeepL.com/Translator (free version)

gurkerl83:
All done, can you please review again, thx!

nacx:
So, with the last update, this PR reverts back to the official gson library, which makes jclouds incompatible with OSGi.

As mentioned in the different JIRA issues, we need to make a decision and find the balance between the cost of supporting OSGi in jclouds itself and being able to move forward faster. In my opinion, if we lack the OSGi expertise to do a proper job there, we should try to stop owning that and delegate that to end-users or to another community that can do it properly.

In any case, this deserves a proper discussion in the jclouds dev/users mailing lists so the community is aware of the impact of the change. It would be great too, to trigger such a discussion cross-posting to jclouds-dev and karaf-dev. The Karaf community kindly took over the jclouds-karaf project and is now maintaining the Karaf and some OSGi bits. I think it makes a lot of sense to involve them in this topic and see also what is in their opinion the best way to keep OSGi support in jclouds from now on.

I think reverting back to official gson should be put on hold until we have clarity on this, and we have a proper deprecation path and plan to make the community aware of the change.

gurkerl83:
The integration of OSGi into JClouds is really only superficial. Not OSGi near component models like Declarative Services or Blueprint were chosen as CDI but simply Guice.

Especially good is already the use of the Service Provider Interface (SPI) as a plugin framework, which is not OSGi specific and has been supported by Java for a long time. 

The OSGi package contained in the core module and its classes, especially MetadataBundleListener through the implementation of the interface BundleListener react when a bundle with an entry in /META-INF/services/ is loaded into the OSGi runtime environment.

In short, you can move the bundle to a new module and the function will be preserved without problems. At this point you have to integrate e.g. Aries SPI-fly bundle into the OSGi runtime environment so that the bundle listener reacts to this kind of events.

Then there is only the part included in ScriptBuilder module, which I will have a look at tomorrow.

Over the weekend I will describe the procedure, which components are affected and how a solution could look like as a JIRA ticket.

I'm also willing to help out with the JClouds aspects handed over to Karaf. I have opened an issue in this project to learn more about the future development of the project.

gaul:
> In any case, this deserves a proper discussion in the jclouds dev/users mailing lists so the community is aware of the impact of the change. It would be great too, to trigger such a discussion cross-posting to jclouds-dev and karaf-dev. The Karaf community kindly took over the jclouds-karaf project and is now maintaining the Karaf and some OSGi bits. I think it makes a lot of sense to involve them in this topic and see also what is in their opinion the best way to keep OSGi support in jclouds from now on.

@gurkerl83 I support this PR but we should give the community a chance to respond to a breaking change.  I don't have the background to explain the tradeoffs with the current OSGi support; would you mind leading this?

nacx:
As @gaul said, I'm not against this PR either, but it is a breaking change that can have significant impact. We know other projects are using our OSGi/Karaf support (such as Apache Brooklyn) and we need to properly communicate and involve the community before breaking it.

core/pom.xml
```diff
@@ -41,9 +41,9 @@
     <jclouds.osgi.import>*</jclouds.osgi.import>
     <!-- https://issues.apache.org/jira/browse/JCLOUDS-1529 -->
     <jclouds.osgi.export>
-      !org.jclouds.json.gson.internal*,
       org.jclouds*;version=${project.version};-noimport:=true
     </jclouds.osgi.export>
+    <jclouds.osgi.include>lib/gson-${gson.version}.jar=gson-${gson.version}.jar;lib:=true</jclouds.osgi.include>
```
gurkerl83, line 46: Instrument maven bundle plugin to integrate the GSON library - Step 1

project/pom.xml
```diff
@@ -641,6 +627,7 @@
             <Import-Package>${guava.osgi.import},${jclouds.osgi.import}</Import-Package>
             <DynamicImport-Package>${jclouds.osgi.dynamic}</DynamicImport-Package>
             <Fragment-Host>${jclouds.osgi.fragment}</Fragment-Host>
+            <Include-Resource>${jclouds.osgi.include}</Include-Resource>
```
gurkerl83, line 630: Instrument maven bundle plugin to integrate the GSON library - Step 2

gurkerl83:
Integrate GSON library in Clouds Core Bundle
The change contained in the commit puts the GSON library into the classpath of the JClouds core module.
After several tests with Karaf and Karaf JClouds, especially if the Maven identifier matches the original GSON library, there are only a limited number of ways to keep the deployment effort low.

Specifically, Karaf has a set of predefined Maven repositories that can be easily customized. The order in which a particular repository is resolved into the customized GSON library is more difficult. In normal OSGi applications, which do not have such a management function, I imagine this configuration to be more complicated.  Sure, a unique identifier would help, but then we are back to step 1.

Although I honestly don't like to see this kind of approach in a codebase I'm working with, there are not many alternatives to the main aspect of deployment mentioned above.

Maybe the approach can still ease the problem in the short term. In a further mid-term step, however, this problem must be addressed in general.

gurkerl83:
Hi, I would love some feedback about latest changes made in the last commit about the handling of GSON, the provided approach and the change of strategy of maintaining OSGi metadata in general. Please look at the last commit the following description.

https://github.com/apache/jclouds/pull/78/commits/0a12d8ff9d4d8d4afda441e0ccfe2e769004628a

In the last commit (last section of squashed commit), the GSON library was integrated into the JClouds core module using maven-bundle plugins include resource instruction. Building OSGi instruction variables from the respective modules show a weakness when resources such as script builder shell scripts are required to be integrated into the bundle but not provide a dedicated variable declaration for the resource section.

The following commit demonstrates a change in strategy in declaration and integration of OSGi metadata.

- Replace old bundle-plugin with newest bnd-plugin (bundle-plugin uses bnd-plugin internally)
- Move OSGi metadata declarations from a maven variable passing strategy into dedicated bnd.bnd files
+ Cleaner pom files, no bundle packaging
+ Intellisense / Autocomplete support for .bnd files in terms of package exports etc.

For demonstration, the overall OSGi adjustments are limited to project, core, script builder, compute, blob store, and load balancer because most custom OSGi metadata is defined here.

Note: Other modules are currently disabled from build because some feedback is needed first.

Make GSON integration work.
To understand the changes, see the core modules' bnd file. GSON internal packages also define a version. Both already exported and new export declarations are fused. The global JClouds core module exports defined the entire set of GSON packages available.

Some minor modifications were made in the module project; replace maven jar plugin with a minified version of the declaration, outsourced in projects bnd file.

nacx:
So, the idea is that the jclouds-core OSGi module can export the son package we want instead of having them directly exported by the gson dependency itself? As long as that is enough to let users use jclouds in OSGI, I think this approach is good and cleaner than the current one.

Have you been able to try this in an OSGi environment? I guess this is still compatible with jclouds-karaf? Or are there changes to be taken into account there?

Thank you very much for taking the time to do the change in a way that facilitates the review; it is much appreciated!.

gurkerl83:
> So, the idea is that the JClouds-core OSGi module can export the GSON package we want instead of directly shipped by the GSON dependency itself?

Yes!

> Have you been able to try this in an OSGi environment?

I have tested loading the core feature in both OSGi environments Apache Felix and Equinox. I was running Apache Karaf with both settings.

> I guess this is still compatible with Jclouds-Karaf, or are there changes to be taken into account there?

According to the Karaf feature description for the core module, a dedicated GSON library gets executed. I guess this instruction corresponds to an older version of the core module, where GSON was a standalone dependency.

The core package also resolves even when a secondary GSON dependency gets loaded.

```xml
<feature name='jclouds' description='jclouds' version='${project.version}'>
    <feature version='${project.version}'>jclouds-guice</feature>
    <bundle dependency="true">mvn:javax.ws.rs/javax.ws.rs-api/${javax.ws.rs.api.version}</bundle>
    <bundle dependency='true'>mvn:com.google.guava/guava/${guava.version}</bundle>
    <bundle dependency='true'>mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.oauth-commons/${net.oauth.bundle.version}</bundle>
    <bundle dependency='true'>mvn:org.apache.servicemix.bundles/org.apache.servicemix.bundles.jersey-core/${jersey.bundle.version}</bundle>
    <bundle dependency='true'>mvn:com.google.code.gson/gson/${gson.version}</bundle>
    <bundle>mvn:org.apache.jclouds/jclouds-core/${jclouds.version}</bundle>
</feature>
```
https://github.com/apache/karaf-jclouds/blob/c117ed392d624f825488f2738feee433fca4d04d/feature/src/main/feature/feature.xml

**Future compatibility with JClouds-Karaf**
The problem with handwritten features like the core module is that the composition of module/feature and its dependencies gets not recomputed automatically. It is not enough to update just the dependency versions. 

I have started a major refactor of Jclouds-Karaf to recalculate each stable and lab feature based on the current dependency tree. I use Karaf Maven Plugin's goal generate-descriptor.

gaul:
@gurkerl83 @nacx Do we have a path forward on this PR?  I would like to include it in the next release.

gurkerl83:
Thank you for checking the status of the Feature Branch, which has been pushed into the background a bit because of other things.

The Feature Branch is not complete, the missing things are in the shelf on my machine. Before I integrate the missing parts, I need your OK on the selected procedure. The most important points are:

- Configuration files are loaded from the outside, written in BND format, replacement of the Maven Bundle plugin used.
- One configuration file per Maven module

Thx

geomacy:
hi all, just regarding Ignasi's [comment](https://github.com/apache/jclouds/pull/78#issuecomment-650507931) and Alex's [concern](https://lists.apache.org/thread.html/r0c1f7bf4c58e6447d145245c8d24de8c05b7787f2b0efb44459bba9a%40%3Cdev.brooklyn.apache.org%3E) on the mailing list, is this still potentially a breaking change for us downstream users (in Apache Brooklyn in this case)? I am hoping based on @gurkerl83's [comment](https://github.com/apache/jclouds/pull/78#issuecomment-652886362) above that this is no longer such a cause for concern?

gurkerl83:
@geomacy My assessment of whether the change is a critical change is as follows. 

The main area of concern is the way Apache JClouds is used within Apache Brooklyn?

At compile time, runtime or both?

If the integration is done at compile time, or running tests, there are no changes.

Runtime
No change during normal JClouds startup. However, if Apache Brooklyn and JClouds are run in an OSGi runtime environment, and Apache Karaf in particular, there may be some kind of incompatibility.

I did a little digging in Apache Brooklyn and found references in one of the git submodules, that Apache Karaf-JClouds features are loaded and started.

Here is the reference to the corresponding file

https://github.com/apache/brooklyn-server/blob/ba7fb9112623872de18bd7b6135fdfc1014228ae/karaf/features/src/main/feature/feature.xml

Per se this is no problem, but Apache Karaf-Jclouds project is on a very old state, which does not map dependencies of newer Apache JClouds versions.

The stack and libraries used in Apache JClouds change over time; consequently, the feature set provided by Apache Karaf-JClouds becomes incompatible over time. The creation of Apache Karaf features is mostly a manual process; it simply does not keep runtime dependencies up to date.

Possible solution
When I made the GSON handling changes in June, I created an automatic feature generator and aligned the outdated Apache Karaf-JClouds project with it. If I find the time in one of the coming weeks I will continue the approach and bring it to finalization, a manual adjustment of features especially in a project with extensive dependencies must be done automatically in the future.

The detection and resolution of possible incompatibilities must consider the launch of Apache Brooklyn, whether each of the outdated feature set of Apache Karaf-JClouds can be resolved.

Thx!

gurkerl83:
The work of this branch is ready for review. 

There is an additional feature in the Apache JCloud Lab project waiting which syncs the OSGi handling due to the shared Maven Parent used in both projects. This branch is also ready for review. 

https://github.com/apache/jclouds-labs/pull/5

Thx!

gurkerl83:
Should I do a rebase and squash?

ahgittin:
Seems to me like this PR is introducing two significant changes:

* use `bnd` instead of `jclouds.osgi.{im,ex}port` maven properties
* change how `gson` is imported/exported

If it's possible to have a PR for the first of these to be discussed, and then a second for gson, that will make it much easier to consider.

The former (`bnd`) seems like a reasonable idea to me.  It's messy either way but slightly less messy.

The latter (`gson` change) is still confusing to me and as per the mailing list my gut says we're better of not touching it unless we are fixing it properly such as by moving to Jackson.

geomacy:
> Should I do a rebase and squash?

Just a suggestion, I think it's easier to follow the history and comments if it is just left as-is.  But I don't know what the conventions are in jclouds.  Perhaps a squash once everything is approved?

Regarding questions [above](https://github.com/apache/jclouds/pull/78#issuecomment-711058485)

> The main area of concern is the way Apache JClouds is used within Apache Brooklyn?
> At compile time, runtime or both?

I would say both. I would share Alex's [concern](https://lists.apache.org/thread.html/r0c1f7bf4c58e6447d145245c8d24de8c05b7787f2b0efb44459bba9a%40%3Cdev.jclouds.apache.org%3E) that Gson 2.8.6 may cause us problems, particularly because of the [issues](https://lists.apache.org/thread.html/r64449ba3747ed3a400725f4fa5bb107680aad067bf13a236253beaf1%40%3Cdev.jclouds.apache.org%3E) Robert points out. Also it seems, iiuc, that if this gets merged then there would be downstream changes required in jclouds-karaf, which we would then need to integrate in turn. Your suggestion of automatic feature generation sounds interesting but would require considerable further work, wouldn't it?

gaul:
I would like to push this review forward.  While I do not understand all the OSGi issues, if they can be resolved by jclouds-karaf making code changes then this should not block the main jclouds code evolution.  Regarding gson 2.8.6 specifically, it seems that google/gson#1601 would block a lot of users?  Would using 2.8.5 temporarily be OK?

@gurkerl83 It is fine to have multiple commits merge in a single PR.  If squashing these makes it easier for a future contributor then please do so, but include the detailed description that you already wrote!

@nacx what do you think?

nacx:
So, the general idea of this is to move to the BND plugin (which I'm OK with) and to address the issue of GSON unexported packages by exporting them from our jclouds-core module. That's OK for me as well.

jclouds needs to be able to upgrade to newer gson versions, and this is the cleanest approach we've had so far to achieve that. If restructuring needed to accommodate the change imposes some changes in downstream projects such as karaf-jclouds integration or Brooklyn, then let it be and get those fixed as well. I think we have been very careful of not breaking stuff and bringing everyone to the same page, but we can't be blocked forever on this and pinned to the old gson library.

@gurkerl83 has already stepped in and started to move forward with the modernization of the karaf-jclouds integration, which is great. I'd suggest we move forward with this PR to unblock jclouds, and anyone interested in having the Karaf and Brooklyn integration aligned with newer jclouds versions to step in and help @gurkerl83 make that happen.

gurkerl83:
Thanks for the feedback!
 
I will leave this feature as it is. I suggest we do not do any squashing to see the respective commits' changes as they appeared; otherwise, it might be harder to comprehend the steps later.

@ahgittin GSON 2.8.5 vs. 2.8.6 - In the OSGi metadata configuration of version 2.8.6, the field require-capability gets generated by the JDK used in the compilation of the library. On the GSON master branch, this field gets predefined in the bnd file to a lower version than version 9, so it can not be overwritten in the library's compilation; only the forthcoming version 2.8.7 works that way. 

In the JClouds Core module, the original GSON library gets embedded (with the original / wrongly configured OSGi metadata file). This does not matter. The OSGi runtime does not interpret this embedded library; otherwise, it would see the limited exports we are trying to fix. Only the JClouds core bundle's OSGi metadata gets analyzed, which also generates its required compatibility flag during compilation. Using JDK 8, it produces the following entry, which is fine.

```Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"``` 

My estimation of using a secondary library of GSON in an OSGi stacked application is to pin its version to 2.8.5 in a JDK 8 environment.

Thx!

**Comment by gurkerl83** [comment]
Can someone review, and when everything is fine, merge please. I have more things coming, but those are based on what it’s in here. Thx!

**Comment by geomacy** [comment]
hi @nacx, about 
> jclouds needs to be able to upgrade to newer gson versions, and this is the cleanest approach we've had so far to achieve that. If restructuring needed to accommodate the change imposes some changes in downstream projects such as karaf-jclouds integration or Brooklyn, then let it be and get those fixed as well. I think we have been very careful of not breaking stuff and bringing everyone to the same page, but we can't be blocked forever on this and pinned to the old gson library.

I don't disagree with this, and I do think this PR is a great bit of work that looks like a good step forward for jclouds. I'm not opposing it, just trying to understand the implications for downstream, and what issues it might cause. 

@gurkerl83 I didn't really understand your response to @ahgittin above about 2.8.5 vs 2.8.6, especially
> My estimation of using a secondary library of GSON in an OSGi stacked application is to pin its version to 2.8.5 in a JDK 8 environment.

With this PR, as you showed [above](https://github.com/apache/jclouds/pull/78#issuecomment-652886362), jclouds will export the gson APIs, at v2.8.6. I can't understand how your users could then use gson at any different version [1]. 

A question for everyone - about the known problems with 2.8.6: what do you think of Andrew @gaul's suggestion [above](https://github.com/apache/jclouds/pull/78#issuecomment-714821651), of staying with 2.8.5 for now? Are there any problems with it that you want to get away from? Or capabilities introduced by 2.8.6 that you want to use in the near future?

In particular, do I understand it right that [google/gson#1601](https://github.com/google/gson/issues/1601) will force your downstream users to use Java 9+ with jclouds once this PR is merged?  I suppose it's hard to know what proportion of your users are still on older versions of Java?

-----
[1] footnote: well, in an OSGI environment you might have a bundle that could ignore the 2.8.6 from jclouds and use instead, say, 2.8.5, but I expect it could only do that if it didn't itself need to use jclouds.

**Comment by gurkerl83** [comment]
@geomacy Sorry about the misleading wording. You are right, the only thing that is happening is a re-export of GSON packages, nothing more. The footnote of yours in response to my estimation is in a practical environment much harder to validate. I think OSGi implementations are not perfect, especially in edge cases like that.

I changed my opinion about which version of GSON we should use in this feature, to make sure we really introduce no side effects on a JDK level. We should go with 2.8.5. Anyway, I would like to introduce a conditional build profile to support JDK 11 and above, here we could then easily set GSON to the 2.8.6 version.

We are not at the finished line, yet. During my latest experiments with the latest version JClouds from the master branch, we might get a problem with the current Guice of version 4.2.3 in use.

https://github.com/apache/jclouds/commit/31a3e5b5df1543d04098e3a694130b7ae8e6e079

Guice 4.2.3 requires a minimal Guava version of 27.1-jre and this is also what my OSGi feature builder is complaining about.
https://mvnrepository.com/artifact/com.google.inject/guice/4.2.3

Luckily the JClouds sources are already compatible with at least guava 29 when the comment in the commit is correct. 
https://github.com/apache/jclouds/commit/62767a14610fc1c97c440dbd5ee0f02b276a1069

I think we have to address this in an additional, much smaller feature, I will open for discussion tomorrow, maybe a JIRA ticket is required also.

Thx!

**Comment by gurkerl83** [comment]
Hi, to continue with other refactorings, I need one of the repository maintainers to review and merge this feature. Please answer as soon as possible, otherwise, I have to delay those work and spend my time on other projects, thx!

**Comment by gaul** [comment]
Thank you for your contribution @gurkerl83!  Note that the jclouds-labs repository may require changes to track this PR.  Could you investigate this?

We should also better understand the compatibility between Guice and Guava.  The intention is to support versions 22-29.  If this is not the case, can you open an issue for tracking?

Thanks everyone for your feedback and patience.  Dependencies are a tough issue for our downstream users and we would appreciate further feedback via Apache JIRA.  I hope to sort out our dependencies and release 2.3.0 later this year.

**Comment by gurkerl83** [comment]
@gaul About JClouds-Lab, there is an open merge request I provided, also ready for merge, which syncs both the new OSGi and GSON approach. The feature is straightforward; please read the commit message because a relocation of cloudsigma packages was necessary. To continue, this feature is also relevant but not so pressure because the jclouds to karaf integration requires both; take your time to review.
About Guice and Guava Binding, Guice 4.2.3 expects in an OSGi runtime Guava with an exact version of 27.1-jre. For now, higher versions like 28, 29, and even 30 are not possible. I had tested this yesterday. 
Thx for merging!

**Comment by nacx** [comment]
Thank you very much for your contribution and patience @gurkerl83! This has been a very much appreciated PR. Thanks for your continued efforts on this!

**Comment by geomacy** [comment]
Thanks for your engagement with our concerns @gurkerl83. I think this will be less likely to cause downstream OSGI problems for jclouds-karaf and/or Apache Brooklyn now that it is using 2.8.5. Your PR looks like a good step forward for jclouds!

**Comment by gaul** [comment]
@gurkerl83 Could you investigate this build failure:

https://ci-builds.apache.org/job/JClouds/job/jclouds-labs-aws/7/ 

I think the changes you made to jclouds-labs need to apply to jclouds-labs-aws and jclouds-labs-openstack as well.

**Comment by gurkerl83** [comment]
@gaul The build error is related to the chances. I make the required changes, didn’t know the project’s existed, respectively get pulled into the build. Thx!

**Comment by gaul** [comment]
Yeah sorry I forgot to mention this.  All the labs variants was a poor engineering decision in retrospect that we will hopefully unwind some day...

**Comment by gurkerl83** [comment]
@gaul I provided two merge requests, one for the aws, and the other for the openstack lab project. The changes should be straightforward, it just aligns the OSGi handling we do in JClouds now in general.

**Comment by ahgittin** [comment]
@gurkerl83 @gaul @nacx @geomacy This breaks any downstream consumers (such as Apache Brooklyn) who don't use the _exact_ version of Guava that jclouds uses.  Guava is part of the public API of the project (methods which take functions etc) so it is important to support a range there.  (Also Guava seems to keep having CVE's against it so it is frequently version-bumped.)

Previously a range of Guava (and okio) packages were allowed, so this is a regression.  I note the `TODO` comments in 7b5c2ae529bee0ed3528df9bc479edde46d9695f about this.  

```
    <guava.version>22.0</guava.version>
    <!-- TODO: this has to be addressed in another feature -->
    <guava.osgi.import>com.google.common.*;version="[22.0,30.0.0)"</guava.osgi.import>

    <guice.version>4.2.3</guice.version>
    <gson.version>2.8.5</gson.version>

    <okhttp.version>2.2.0</okhttp.version>
    <!-- TODO: this has to be addressed in another feature -->
    <okio.osgi.import>okio;version="[1.2.0,1.3)"</okio.osgi.import>
```

This was introduced by the switch from `maven-jar-plugin` to `bnd-maven-plugin` in 8ac994c2b5d509ca2877efdd77c35881d58ac3e3 as part of this PR.  Obviously the switch to bnd means the import-package range instruction needs to be configured in a different way.

Is that in the works?

We are blocked from using newer jclouds until this is resolved, as is anyone trying to use it in an OSGi context unless they can use the exact Guava version.

gaul:
@ahgittin I feared that this would impact downstream consumers.  Unfortunately our team lacks the resources to maintain OSGi compatibility.  We will happily accepts PRs that plausibly improve compatibility, as this one appears to.

geomacy:
@gurkerl83 what did you have in mind when you added that TODO comment in 7b5c2ae? Do you know of a way to support a range of package-import versions in downstream projects using the bnd mechanism?

gurkerl83:
Hi all,

The Guava version fixed to 22.0 in the master branch was already set before the refactoring performed in this branch. 

Different versions of Guava were allowed in an OSGi environment by overriding the 22 version (derived from the Maven dependencies) with a range of 22.0 -> 30.0.0.

In the present feature, the two properties 
- guava.osgi.import and
- okhttp.version
have not been attached to the build process (Bnd plugin), but have been recorded as comments only.

Before the resolution method is considered, the problem of Guava and its close coupling with Guice must be examined.

The Guice project possesses, differently to GSON no explicit configuration, how its dependencies are regarded in an OSGi execution environment.

Guice / Guava
4.1.0 / 19.0
4.2.0 /  [23.6,24)
4.2.1 / [25.1,26)
4.2.3 / [27.1,28) => used
5.0.0.BETA-1 / [27.1,28)

The version of Guice built into JClouds is 4.2.3. Increasing to this version number was not the subject of this branch.

To avoid backward compatibility problems between Guice and Guava (e.g. Guice 4.2.3 -> Guava 24) this means that the version of Guava integrated in JClouds must be at least version 27.1. Guice puts this in its dependencies. 

In my opinion, consumers (JClouds, as well as consumers of JClouds) should take this minimum barrier into account.

Resolution
To understand the resolution, please consider the following previously unnoticed branch.

In order to add a Guava verison range to the bundles rolled out by JClouds, the following import rule is added in the BND Configuration of the Project Module.

Import Package: \
    com.google.common.*;version="[27.1,30.1.0)", \
    okio.*;version="[1.2.0,1.3)", \
    *

In the output of each JClouds module that uses Guava (and there are some) the configuration to the range version is observed. The bnd built into the JClouds project modules acts as a base of other bnd configurations, and should also apply in the Lab and Co repositories. Adjusted artefacts only become valid when a release is available or the build of these repositories is triggered again.

Output - Core Module
Import package:
   com.google.common.base;version="[27.1,30.1.0)",
   com.google.common.cache;version="[27.1,30.1.0)",
   com.google.common.collect;version="[27.1,30.1.0)",
   ...
   com.google.gson;version="[2.8,3)",
   ...

Output - Blobstore Module
Import package:
   com.google.common.base;version="[27.1,30.1.0)",
   com.google.common.collect;version="[27.1,30.1.0)",
   com.google.common.hash;version="[27.1,30.1.0)",
   ...
   com.google.inject;version="[1.4,2)",
   ...

Oki version range
The okhttp dependency requires okio version 1.2. The ok* libraries respectively the dependencies used in these libraries do not have OSGi instructions. For this reason, the previously integrated configuration is used. 

Local tests show that the import directive can also be set globally, see above. An explicit specification in the BND file of the JClouds okHTTP module is not useful because this JClouds module also uses Guava.


Output - OkHTTP Module
Import package:
   com.google.common.base;version="[27.1,30.1.0)",
   com.google.common.collect;version="[27.1,30.1.0)",
   okio;version="[1.2.0,1.3)",
   com.google.inject;version="[1.4,2)",
   ...

Note: In the following feature, the version of the ok* framework is heard. Extensive adjustments are also performed.

At this point, it should be noted that the import policy in the okio will probably have to be extracted again because standalone OSGi metadata may be provided.  

Translated with www.DeepL.com/Translator (free version)

geomacy:
Thanks for the extensive notes @gurkerl83. I'm not sure what that means we need to do, but we can continue the conversation in #86 (or perhaps simply on the jclouds mailing list) as you suggest.

gurkerl83:
@geomacy Already done in the commit https://github.com/apache/jclouds/pull/86/commits/625e54d2028a456d15bd142fda77196019ce6f3f

gaul:
Is there anything we can do to improve this in the short-term?  I would like to release jclouds 2.3.0.

gurkerl83:
@gaul What do you mean, is the min version of guava, which is set to 27.1 to high? If everyone is fine please go ahead and review the mentioned feature. A release shouldn’t be a problem then, thx!

gaul:
`project/pom.xml` shows Guava the minimum version as 22.0 and the OSGi version range as [22.0,30.0.0).

gurkerl83:
@gaul Those things are cleaned up in the following open merge request https://github.com/project-millipede/jclouds/tree/Maven-Configuration-Cleanup Please review and merge, thx

PR #79, opened 2020-06-26
JCLOUDS-1542 follow-up: check whether isAccessible() is already set

JnRouvignac:
`AccessibleObject.canAccess()` would be much much better,
but this API can only be used from Java 9 onward.

Note: I have not tested this code, I will see if we can test it on Monday.
I have opened the PR to get the review going and see if anything needs to be changed.
Thank you!


nacx:
Thanks! :)

PR #82, opened 2020-09-11
JCLOUDS-1552: Do not attempt to parse empty payload

cstamas:
In  case of empty response, AWSUtil returns null, but produces WARN logspam.

https://issues.apache.org/jira/browse/JCLOUDS-1552

### DIFF
--- a/apis/sts/src/test/java/org/jclouds/aws/util/AWSUtilsTest.java
+++ b/apis/sts/src/test/java/org/jclouds/aws/util/AWSUtilsTest.java
@@ -17,8 +17,13 @@
 package org.jclouds.aws.util;
 import static javax.ws.rs.core.MediaType.TEXT_PLAIN;
 import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
+import static javax.ws.rs.core.Response.Status.NOT_FOUND;
+import static org.easymock.EasyMock.anyObject;
+import static org.easymock.EasyMock.anyString;
 import static org.easymock.EasyMock.createMock;
 import static org.easymock.EasyMock.expect;
+import static org.easymock.EasyMock.expectLastCall;
+import static org.easymock.EasyMock.mock;
 import static org.easymock.EasyMock.replay;
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.assertNull;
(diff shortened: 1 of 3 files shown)


apis/sts/src/test/java/org/jclouds/aws/util/AWSUtilsTest.java
```diff
@@ -88,6 +95,20 @@ public void testNoExceptionParsingTextPlain() {
...
+   @Test
+   public void testNoExceptionEmptyPayload() {
+      utils.logger = mock(Logger.class);
+      utils.logger.warn(anyObject(Throwable.class), anyString());
+      expectLastCall().andThrow(new IllegalStateException("log spam"));
+      replay(utils.logger);
+      HttpResponse response = HttpResponse.builder().statusCode(NOT_FOUND.getStatusCode()).payload(new StringPayload("")).build();
+      response.getPayload().getContentMetadata().setContentType("application/unknown");
+      assertNull(utils.parseAWSErrorFromContent(command.getCurrentRequest(), response));
+   }
```
nacx, line 110: Is there any way we can better test this? The whole purpose of the change shouldn't be to _not send logs_, but to avoid parsing a body that we know is empty.
Instead of mocking the logger, could we for example mock the response.getPayload to make sure it is not called more than once? Or something similar that helps us verify that the code reached the desired execution paths, without having to mess up with the logger.

nacx (commented):
Thanks! LGTM. Just one comment about the approach used in the test.

apis/sts/src/test/java/org/jclouds/aws/util/AWSUtilsTest.java
cstamas: Hey, sorry for late response. Well yes, that could be made as well, but...

I really wanted to make sure that execution does not end up in `catch` branch, and there logger is the only thing I could get hold of :) I mean, in future the code may change (for example to reuse the `parseAWSErrorFromContent` byte[]), and then the mocked test would may become "false negative". This way, am sure my goal is achieve: there is no WARN log spam :smile: 

But in general, am open to both changes.

nacx: So, I think the right way to approach this is a bit different. You don't wanna check that the code does not reach the `catch` branch. You wanna verify that it does not try to even parse the response payload. Having a try/catch there is anecdotical; the logic you wanna test is that if the payload is present but empty, the code doesn't attempt to parse it.

For that, looking at the code you could for example mock the factory that is used to get the apply function, or even provide a mock apply function and fail if it is called. I think this approach would much better test the behaviour we want to fix here. WDYT?

cstamas: +1 will try to get it later afternoon

nacx (approved):
Looks much cleaner, thanks!

nacx:
Cherry-picked to 2.2.x as well.

PR #84, opened 2020-10-09
JCLOUDS-1551: Update version of OkHttp 3.14.9

korlov42:


gaul:
@korlov42 Can you look at the CI failure?

```
Caused by: java.lang.UnsupportedOperationException: clientBuilder.sslSocketFactory(SSLSocketFactory) not supported on JDK 9+
	at okhttp3.internal.platform.Jdk9Platform.trustManager(Jdk9Platform.java:85)
	at okhttp3.internal.platform.Platform.buildCertificateChainCleaner(Platform.java:177)
	at okhttp3.OkHttpClient$Builder.sslSocketFactory(OkHttpClient.java:743)
	at org.jclouds.http.okhttp.config.OkHttpCommandExecutorServiceModule$OkHttpClientProvider.get(OkHttpCommandExecutorServiceModule.java:85)
	at org.jclouds.http.okhttp.config.OkHttpCommandExecutorServiceModule$OkHttpClientProvider.get(OkHttpCommandExecutorServiceModule.java:56)
```

nacx:
I remember quite some time ago I had a look at OkHttp3, but found that some of the APIs some providers have conflicted with validations in OkHttp3, such as rejecting DELETE operations that have a payload.
It's been a while but, do those validations still exist? Make sure that is taken care of as well. I don't remember if it affected just providers in `jclouds-labs` or to fully-supported providers. If that's still an issue we may want to use this to discuss how to fix that.

korlov42:
@nacx, for now, only GET and HEAD requests will be rejected if they have a body. But there is another validation that forced me to slightly modified the production code: rejecting requests without a body when the body is required (POST, PUT, PATCH and a few more). So, I've decided to rely on the internal method of okhttp (HttpMethod#requiresRequestBody) and generate an empty request body. You could see it [here](https://github.com/apache/jclouds/pull/84/files#diff-0669ed207d502fa8adc5114f1c4e8167R100).

gaul:
This now fails with a compile error:

```
/home/travis/build/apache/jclouds/providers/profitbricks/src/test/java/org/jclouds/profitbricks/features/ImageApiMockTest.java:20: error: package com.squareup.okhttp.mockwebserver does not exist

import com.squareup.okhttp.mockwebserver.MockWebServer;
```

You might want to run `mvn verify` locally.

korlov42:
The IDEA decided to put changes of one particular file to another change list, thus locally there is no problem, but these changes have missed in remote repo.

gaul:
@korlov42 Thank you for taking on this large task!  The change looks good.  How can we best test this so that we have confidence that it works?  Have you tried running against a real provider when overriding the transport to OkHttp?  Also if we merge this, will you also fix any of the necessary jclouds-lab code?

@nacx Any other suggestions?  I am looking forward to upgrading my local JDK when running jclouds tests!

korlov42:
> Have you tried running against a real provider when overriding the transport to OkHttp?

Unfortunately, I don't have possibility to test this on real-world environment, so I've relied only on those tests presented in project.

> Will you also fix any of the necessary jclouds-lab code?

Maybe yes, I could help with that.

nacx:
>for now, only GET and HEAD requests will be rejected if they have a body. But there is another validation that forced me to slightly modified the production code: rejecting requests without a body when the body is required (POST, PUT, PATCH and a few more)

Could you add some logs in the driver that would print a warning when invoking methods that are not expected to have a body with a body, and when invoking methods that require a body without it? This way users could easily spot potentially problematic APIs and we could have a quick overview of the providers that use OkHttp by default and use those unexpected APIs.

gaul:
@korlov42 We plan to release 2.3.0 later this month.  Do you think you can finish this PR and the jclouds-lab equivalent soon?

gaul:
@korlov42 any chance you could rebase and revisit this PR?

gurkerl83:
@gaul @korlov42 I have rebased the branch and made a pull request here https://github.com/apache/jclouds/pull/95. I made the respective changes in labs, lab-aws, and lab-opencore. The build of all those repos is fine, all tests green. I would like to continue the discussion over there because of a few things I have noted, especially with the exclusion of bouncycastle in mockwebserver and the OSGi handling stuff of OKHttp, OKTLS, OKIO.

gurkerl83:
@gaul A few days ago I made all required changes. Please see all available merge requests first because the current merge request is outdated. Thx!

gaul (approved):
Sorry wrong issue, closing in favor of #95.

PR #86, opened 2020-10-26
Initial cleanup of maven set up in the parents' module

gurkerl83:
- Increase version of Guava dependency from 22.0 to 27.1-jre.

Note: In an OSGi runtime, Guice version 4.2.3 expects Guava with an exact version of 27.1-jre. For now, higher versions of Guava are not possible.

- Separate version numbers from dependencies and plugins

To understand the scope of the modification work performed in the feature branch, I refer you to the notes listed in the individual commits.

**Due to an approaching release, these optional modifications will be considered in the upcoming development circle.**

https://issues.apache.org/jira/browse/JCLOUDS-1555

~~Following is a description of the purpose of the feature.~~

~~New contributors find it difficult to understand the function and intent of different plugins integrated into JClouds.~~

~~The primary goal is to simplify and streamline the overall setup.~~

~~On closer examination of the integrated plugins you will notice that,~~

~~- different plugins are used for the same or a similar goal~~
~~- the development activity of used plugins has been terminated~~
~~- modified integration requirements through Maven. The development of used plugins is progressing. Versions of the plugins integrated into JClouds are not up to date. Newer plugin versions require an adjustment to how integration into a project is done.~~

~~Basically the setup of the plugins used in the Maven project POM was done a long time ago. The project POM file of the JClouds project serves as a parent for all modules in JClouds.~~

~~The first step is to consolidate the used plugins.~~

~~The same aspects apply to the build profiles used in the project. On closer look these must be considered obsolete. A deconstruction of declared build profiles is aimed at.~~

~~In order to support newer runtime environments, for example JDK11 and above, we aim to integrate build profiles.~~

~~In the respective build profile the libraries which are no longer contained in the respective target JDK but needed in JClouds are to be integrated.~~

~~In the course of this rebuild, a property-based way of declaring the version number of dependencies and plugins will be implemented.~~

### DIFF
--- a/apis/chef/pom.xml
+++ b/apis/chef/pom.xml
@@ -63,11 +63,6 @@
       <type>test-jar</type>
       <scope>test</scope>
     </dependency>
-    <!--  for ohai -->
-    <dependency>
-      <groupId>com.google.inject.extensions</groupId>
-      <artifactId>guice-multibindings</artifactId>
-    </dependency>
     <!--  for transient chef provider -->
     <dependency>
       <groupId>org.apache.jclouds</groupId>
--- a/apis/docker/pom.xml
+++ b/apis/docker/pom.xml
@@ -79,10 +79,6 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
-      <groupId>com.google.inject.extensions</groupId>
-      <artifactId>guice-multibindings</artifactId>
-    </dependency>
-    <dependency>
       <groupId>org.apache.jclouds</groupId>
       <artifactId>jclouds-core</artifactId>
       <version>${project.version}</version>
--- a/apis/openstack-keystone/pom.xml
+++ b/apis/openstack-keystone/pom.xml
@@ -52,11 +52,6 @@
       <artifactId>jclouds-core</artifactId>
       <version>${project.version}</version>
     </dependency>
-    <!-- for the extension namespaces -->
-    <dependency>
-      <groupId>com.google.inject.extensions</groupId>
-      <artifactId>guice-multibindings</artifactId>
-    </dependency>
     <dependency>
       <groupId>org.apache.jclouds</groupId>
       <artifactId>jclouds-core</artifactId>
--- a/apis/openstack-neutron/pom.xml
+++ b/apis/openstack-neutron/pom.xml
@@ -52,11 +52,6 @@
       <artifactId>jclouds-core</artifactId>
       <version>${project.version}</version>
     </dependency>
-    <!-- for the extension namespaces -->
-    <dependency>
-      <groupId>com.google.inject.extensions</groupId>
-      <artifactId>guice-multibindings</artifactId>
-    </dependency>
     <dependency>
       <groupId>org.apache.jclouds</groupId>
       <artifactId>jclouds-core</artifactId>
--- a/apis/openstack-nova/pom.xml
+++ b/apis/openstack-nova/pom.xml
@@ -54,11 +54,6 @@
       <artifactId>openstack-neutron</artifactId>
       <version>${project.version}</version>
     </dependency>
-    <!-- for the extension namespaces -->
-    <dependency>
-      <groupId>com.google.inject.extensions</groupId>
-      <artifactId>guice-multibindings</artifactId>
-    </dependency>
     <dependency>
       <groupId>org.apache.jclouds</groupId>
       <artifactId>jclouds-compute</artifactId>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -64,7 +64,7 @@
     </dependency>
     <dependency>
       <groupId>org.osgi</groupId>
-      <artifactId>org.osgi.compendium</artifactId>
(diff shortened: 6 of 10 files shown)


gaul:
@gurkerl83 We plan to release 2.3.0 later this month.  Do you think you can finish this PR soon?

gurkerl83:
@gaul Yes, I will do this next week

gurkerl83:
@gaul The feature is basically ready for a review and a second opinion. The most important change in this feature is probably due to a Guice / Guava conflict. The limitation of the version was tested. Please take a look at the individual comments made in the commits.

I suggest to postpone major enhancements, especially regarding the usage of the plugin infrastructure, until the next development cycle, because a discussion about the future behaviour is necessary. 

Thx!

gurkerl83:
@gaul I have a small problem with the maven-site-plugin in the base project pom file. I use IntelliJ and its complaining that report-plugins within the configuration section should be done in the reporting section of the pom. Is this plugin with all its report plugins still in use. One of the plugins used here is java-doc, but java-doc is also in in some doc profiles spread across several poms. Any hint on how to handle this?

geomacy (commented):
I can't comment on the exact details in this but it certainly looks like it is making the dependencies cleaner.

Does the `com.google.common.*;version="[27.1,30.1.0)"` declaration mean that downstream consumers such as Apache Brooklyn will be able to use a range of Guava versions, so that it would address the concern raised by @ahgittin in https://github.com/apache/jclouds/pull/78#issuecomment-753944755 for PR #78?

gurkerl83:
@geomacy 

> Does the com.google.common.*;version="[27.1,30.1.0)" declaration mean that downstream consumers such as Apache Brooklyn will be able to use a range of Guava versions, so that it would address the concern raised by @ahgittin in #78 (comment) for PR #78?

Yes it does! Downstream consumers can use any version between 27.1-jre and 30.1-jre in their projects.

geomacy:
> Yes it does! Downstream consumers can use versions of guava between 27.1-jre and 30.1-jre in their projects.

Marvellous, I am glad to hear it. Thanks!

gaul:
Sorry about my delayed response.  I appreciate breaking out the commits and I follow the logic from the detailed messages.  However I am concerned with upgrading the Guava version:

> Note: In an OSGi runtime, Guice version 4.2.3 expects Guava with an exact version of 27.1-jre. For now, higher versions of Guava are not possible.

A later commit message says:

> The version of Guice built into JClouds is 4.2.3. Increasing to this version number was not the subject of the current branch.
>
> To avoid backward compatibility problems between Guice and Guava (e.g. Guice 4.2.3 -> Guava 24) this means that the version of Guava integrated in JClouds must be at least version 27.1. Guice puts this in its dependencies.

Prior to this PR, jclouds supported Guava 22-30, or at least we believed that this range worked.  Is it possible to continue this support, possibly by downgrading Guice?  I am not opposed to upgrading but I want to understand if there are other options since Guava upgrades have previously caused user complaints and rollbacks.

Also the last message for okio confuses me:

> In the following feature, the version of the "ok" framework has been significantly increased.  Extensive adjustments are also performed.

The version appears to be the same?  We have another PR #84 which attempts to upgrade okio.

We have still not released jclouds 2.3.0 so I am considering including this PR for the upcoming release.  2.3.0 upgrades a bunch of other dependencies, including requiring Java 8, so I think bumping a few more will be OK.  What do you think?

gurkerl83:
@gaul, thanks for some feedback,

about Guice, the current version in master is 4.2.3; previous to this version, Guice 3.0 was used.

Guice 3.0 did not use guava as a dependency; this was introduced in guava 4
https://mvnrepository.com/artifact/com.google.inject/guice/3.0
https://mvnrepository.com/artifact/com.google.inject/guice/4.0

Using a Guice version closer to 4 like 4.1 does not set the boundary of guava to high (guava 19 required) but provides the risk of losing significant improvements made in Guice during 2016 - 2020, which is think is more critical.

> Before this PR, jclouds supported Guava 22-30
This was an estimation, maybe by experiment rather than looking at the possible version ranges in general. Guice 3.0 was used previously; the guava version range was probably meant for a different problem and was not related to Guice. In my opinion, it is acceptable to set some minor requirements a delegate those to the consumer. When you keep this forever, no progress is going to be made.

> Also, the last message for okio confuses me: ... The version appears to be the same?
They are the same. I wrote this comment taking into account forthcoming changes that might get introduced by https://github.com/apache/jclouds/pull/84. Because of no recent activity about this branch's final state, it is unlikely to be considered for a release.

> 2.3.0 upgrades a bunch of other dependencies, including requiring Java 8, so I think bumping a few more will be OK.
What dependencies have you in mind, more like dev dependencies used for the build or those in the code?

gaul:
Let's merge this as-is.  While I have concerns about Guava, this PR is overwhelmingly a step in the right direction.  I will send a message to jclouds-user and see if we can get some testing and feedback on previous and these dependency upgrades.

gaul:
Thank you for your contribution and patience @gurkerl83!

PR #87, opened 2020-11-09
JCLOUDS-1556: Do not parse AWS HEAD response

zhwang-amad:
According to RFC 2616 there cannot be a response body. Trying to parse
it results in a spurious error being logged.


zhwang-amad:
Already fixed by #82 . My bad.

PR #88, opened 2020-11-11
JCLOUDS-1557 - Azure local server support

davidsloan:
Adding ability to use local azure-compatible services, eg Azurite

- Ensuring the endpoint URL is respected by SharedKeyLiteAuthentication and AzureBlobRequestSigner via the new StorageUrlSupplier
- Adding AppendAccountToEndpointModule module to enable the URL format of account key in the path. By default it respects the default Azure behaviour of the storage account name as part of the virtual host.
- Updating unit tests


providers/azureblob/src/main/java/org/jclouds/azure/storage/util/StorageUrlDelegate.java
```diff
@@ -0,0 +1,66 @@
...
+
+@Singleton
+public class StorageUrlDelegate {
+
+   private ProviderMetadata providerMetadata;
+   private Supplier<Credentials> credentialsSupplier;
+   private boolean storageAccountInsideVirtualHost;
+
+   @Inject
+   public StorageUrlDelegate(ProviderMetadata providerMetadata,
```
nacx, line 38: Instead of manually getting the endpoint from the provider metadata, you should inject the following endpoint supplier and get it from there, when needed.
```suggestion
   public StorageUrlDelegate(@Provider Supplier<URI> endpointSupplier,
```

providers/azureblob/src/test/java/org/jclouds/azure/storage/util/StorageUrlDelegateTest.java
```diff
@@ -0,0 +1,83 @@
...
+   void testCustomEndpointWithoutStorageAccountPath() {
+
+      StorageUrlDelegate target = ContextBuilder
+            .newBuilder("azureblob").endpoint("http://localhost:10000")
+            .credentials(ACCOUNT, "?token")
+            .modules(MODULES)
+            .buildInjector()
+            .getInstance(StorageUrlDelegate.class);
+
+      assertEquals(target.configureStorageUrl(), "http://localhost:10000/");
```
nacx, line 61: Why are we only setting the account name if `PROPERTY_AZURE_VIRTUAL_HOST_STORAGE_ACCOUNT == false` is false? I mean, what is the point in not adding the account name anywhere in the URI? Either if we have a custom endpoint or not, I guess we want the account name to be present _somewhere_?

nacx (commented):
Thanks, @davidsloan!

providers/azureblob/src/test/java/org/jclouds/azure/storage/util/StorageUrlDelegateTest.java
davidsloan: For azure, it's already present in the vhost/subdomain, having it in the path too would be incorrect.  Therefore it must be configurable as the behaviour of the azurite endpoint looks for it in the path.

providers/azureblob/src/main/java/org/jclouds/azure/storage/util/StorageUrlDelegate.java
davidsloan: Thanks for the tip, I've implemented the change you suggested!

providers/azureblob/src/test/java/org/jclouds/azure/storage/util/StorageUrlDelegateTest.java
nacx: My point is, does it make sense to _ignore_ the account if a custom endpoint is configured? Should we just do the same as we do when `PROPERTY_AZURE_VIRTUAL_HOST_STORAGE_ACCOUNT == false` when a custom endpoint is set (regardless of the property value)?

Which leads me to think: doe sit make sense to set the property to `vhost` if we're setting a custom endpoint? because the current code does not handle that case. And if it does not make sense, do we still need the property? Can we just default to vhost, but if there is a custom endpoint, just append the account to the path?

davidsloan: Possibly. I was following the pattern used by the AWS S3 connector which has a similar property.

But I'm not sure there's any harm in having it configurable instead of automatic?

nacx: It is about correctness. If there is a custom endpoint and the property has the default value, the account part never manifests in the generated storage URL. Is that correct?

davidsloan: There are use cases where it is be desirable to have a custom endpoint without activating the storage-account-in-path format.

For example you set up a custom DNS name for your blob storage endpoint: 
https://docs.microsoft.com/en-us/azure/storage/blobs/storage-custom-domain-name?tabs=azure-portal

This would not be running the local service but you would be using a custom endpoint.

This is not a use case that I am interested in, but just throwing it out there.

nacx: That makes sense, thanks for the link!

One last thing, then, if there is no custom endpoint configured and the property is set to `false`, the account will be included in the hostname, but appended to the path as well, which I guess is not correct?

davidsloan: No worries, I appreciate the comments.

Sure, it defaults to "true". If the behaviour is explicitly configured to "false" in the scenario you've outlined, then you are right this would not be correct.

I don't think it's possible to prevent misconfigurations like this? I would not expect most users to override the default behaviour unless they had a reason to?

Did you have any suggestions?

nacx: The problem here is that given random endpoints, we can't preconfigure a correct way of setting the account for every possible case. What bout this:

Let's change the `StorageUrlDelegate` to an interface like this:
```java
@ImplementedBy(StorageAccountInVhost.class)
public interface StorageURLSupplier extends Supplier<URI> {
}
```
And have the default implementation be `StorageAccountInVhost`, which just keeps the current behavior, with the hardcoded endpoint and the identity in the host.

Then we could have a:
```java
public class AppendAccountToEndpoint implements StorageURLSupplier
```
That just appends the identity to whatever endpoint is configured.

To make it easier to configure, we could have a Guice module as well that does this (but is not included by default in the azure blob modules list:
```java
public class AppendAccountToEndpointModule extends AbstractModule {
   @Override
   protected void configure() {
      bind(StorageURLSupplier.class).to(AppendAccountToEndpoint.class);
   }
}
```
With these three pieces there, you can plug in your storage URL strategy when creating the context by adding the `AppendAccountToEndpointModule` to the list of modules.

This way we don't need the property, achieve the same behavior, but we also allow users to customize this further, as they can provide their own module to configure how the identity should be added to the endpoint.
This approach may seem overkill, but it ensures correctness in configuration.

WDYT?

gaul (commented):
Please make sure Azure integration tests pass:

```
mvn integration-test -pl :azureblob -Plive -Dtest.azureblob.identity="${JCLOUDS_IDENTITY}" -Dtest.azureblob.credential="${JCLOUDS_CREDENTIAL}"
```

gaul:
@davidsloan We plan to release 2.3.0 later this month.  Do you think you can finish this PR soon?

providers/azureblob/src/test/java/org/jclouds/azure/storage/util/StorageUrlDelegateTest.java
davidsloan: Sounds good, I've refactored the solution to follow your approach.  Thanks.

davidsloan:
> Please make sure Azure integration tests pass:
> 
> ```
> mvn integration-test -pl :azureblob -Plive -Dtest.azureblob.identity="${JCLOUDS_IDENTITY}" -Dtest.azureblob.credential="${JCLOUDS_CREDENTIAL}"
> ```

@gaul There seem to be 4 failures when I run locally, these are present on `master` as well if I run the command you supplied so I don't think they can be related to this change.

- testCopyIfNoneMatchNegative(org.jclouds.azureblob.blobstore.integration.AzureBlobIntegrationLiveTest)
- testPutMultipartInputStream(org.jclouds.azureblob.blobstore.integration.AzureBlobIntegrationLiveTest)
- testCopyBlobIfModifiedSince(org.jclouds.azureblob.AzureBlobClientLiveTest)
- testCopyBlobIfNoneMatch(org.jclouds.azureblob.AzureBlobClientLiveTest) 



> @davidsloan We plan to release 2.3.0 later this month. Do you think you can finish this PR soon?

Hopefully there are no further changes to be made. :+1:

gaul:
Thank you for your contribution and testing @davidsloan!

PR #91, opened 2020-12-06
Maven cleanup

gurkerl83:


gurkerl83:
Somehow the build is stuck, I will push another commit maybe it is restarting...

gurkerl83:
Hi, can you summarize the release process, please, or reference a document that describes it in terms of which plugins and profiles get used? The maven release plugin integrated into the project module file calls a profile "apache-release" defined in the apache parent pom? Thx!

nacx:
You can find the release process here:
https://cwiki.apache.org/confluence/display/JCLOUDS/Releasing+jclouds

We don't use the maven release plugin to update the poms. We do it manually with a custom script. In that link, you'll find the process and the links to the different scripts we use.

gurkerl83:
@nacx Thx for the link.

The build runs without using the following declaration.
```
<parent>
    <groupId>org.apache</groupId>
    <artifactId>apache</artifactId>
    <version>14</version>
    <relativePath />
  </parent>
```

I have a question about the parent declaration; git says it got added in 2013 / 14 before it was an apache project. More modern versions are available. Is this a requirement of an Apache Project using this parent declaration, or is this optional? I have no problem with keeping it for now, but in the future, the maven setup found in the repository should describe the necessary tools and settings for a build. Because maybe the project likes to be compatible with more modern JDKs, new profiles are required, and side effects from upstream projects declarations have to be avoided.
Thx!

gurkerl83:
@gaul this PR is ready for review. It is the first step of a major maven cleanup, maybe we can provide this for the upcoming release.

gaul:
@nacx Any comments on this PR?

nacx:
This LGTM from the peanut gallery, but we need to make sure the Jenkins builds (which we use to run the release) get back to green:
https://ci-builds.apache.org/job/JClouds/

gurkerl83:
@nacx @gaul The jenkin builds for jclouds is green, for lab lab-aws and lab-openstack they are not right now. The reason is the merge order. Okhttp was updated in the project module, and not just the Version but also the groupId. In my opinion the builds are red because the jclouds build was not finished yet, with the new project snapshot artifact when the respective merge of the lab pull request started.

nacx:
The labs builds are triggered automatically after the main one completes, and they're back to green :)

BTW, given that you're here, the [compact build](https://ci-builds.apache.org/job/JClouds/job/jclouds-compat/) is broken now. What versions do we support now, with the recent changes? It would be great to reconfigure that build to make sure we have CI covering our supported version matrix.

gurkerl83:
One commit made in this pull request removed the maven jenkins profile https://github.com/apache/jclouds/pull/91/commits/1d40dd974fe49f85afb0b5407e81b90087f898db. I was pretty sure back then that I saw only src and doc profiles required. 

Looking at the the jclouds and jclouds-compact build now, it seems only jenkins is required also. It seems the Main Build goes through even when one of the profiles, in this case jenkins, is missing. Maybe the change of this commit has to be reverted.

I can not really help with the jenkins ci configuration. I can only provide the version ranges of guice and guava required.

at least >= JDK 1.8 (latest), guava >= 27.1-jre, guice == 4.2.3.
Note: 27.1-jre is really a version number here!

In the compact build the versions of guava and guice are defined from the maven cli. In oder to set the right version the property version names have to be the same, but no change here, just change the values to

-Dguava.version=27.1-jre -Dguice.version=4.2.3

maybe this helps?

nacx:
Ok, I've updated the build matrix to reflect this: https://ci-builds.apache.org/job/JClouds/job/jclouds-compat/

gurkerl83:
It seems the build is only successful with JDK 8. During my work in JClouds I have not compiled it with a higher version locally. Maybe JDK 9 and higher do not work because the JClouds Maven config applies settings which are not available in higher versions anymore.

PR #92, opened 2020-12-14
JCLOUDS-1561: add support for proxy exclusion by URI pattern

roded:


core/src/main/java/org/jclouds/Constants.java
```diff
@@ -182,6 +182,11 @@
     */
    public static final String PROPERTY_PROXY_FOR_SOCKETS = "jclouds.proxy-for-sockets";
 
+   /**
+    * List of URI patterns for which the defined proxy will not be used.
+    */
+   public static final String PROPERTY_PROXY_EXCLUDE_URI_PATTERN_PREFIX = "jclouds.proxy-exclude-uri-pattern.";
```
gaul, line 188: What is the format of this?  If it's a list, is it comma-seperated?

core/src/main/java/org/jclouds/proxy/ProxyExcludePatterns.java
```diff
@@ -0,0 +1,73 @@
...
+ */
+public class ProxyExcludePatterns {
+
+    @Resource
+    protected Logger logger = Logger.NULL;
+
+    private final Set<Pattern> excludeUrisPatterns;
+
+    @Inject
+    ProxyExcludePatterns(Function<Predicate<String>, Map<String, String>> filterStringsBoundByName) {
```
gaul, line 46: This seems like special case behavior that other users are unlikely to discover.  Instead of modifying `ProxyForURI`, could you override `RestModule.configure` in your application?

gaul (changes requested):

core/src/main/java/org/jclouds/Constants.java
roded: I was following the PROPERTY_TIMEOUTS_PREFIX example to allow specifying multiple patterns via multiple configuration options. Using commas and such would've required escaping which I preferred to avoid.

core/src/main/java/org/jclouds/proxy/ProxyExcludePatterns.java
roded: You mean in my application, provide ContextBuilder with my own Guice Module which overrides RestModule and binds a Function<URI, Proxy> of its own (probably reusing ProxyForURI)? I had not considered that option. 
Will Guice not be bothered by the duplicate binding of the Function? Is this the recommended method to alter JClouds' behavior? Thanks.

gaul:
Closing due to narrow use case and inactivity.  Please reopen if you think there is something we can do to make jclouds more extensible without adding configuration knobs that few users will configure.

PR #93, opened 2021-01-06
Simplify S3 code that uses java-xml-builder

JnRouvignac:
- animal sniffer should be on java18, just like `<jdk.version>`
- Only use XMLBuilder's elem() and text() methods to have similar looking code
- Remove unnecessary call to XMLBuilder's up() because the returned value is never used
- Simplify code
- Deduplicate code
- Make the code more explicit by returning the rootBuilder

### DIFF
--- a/apis/s3/src/main/java/org/jclouds/s3/binders/BindACLToXMLPayload.java
+++ b/apis/s3/src/main/java/org/jclouds/s3/binders/BindACLToXMLPayload.java
@@ -59,9 +59,9 @@ public class BindACLToXMLPayload implements Binder {
             S3Constants.S3_REST_API_XML_NAMESPACE);
       if (acl.getOwner() != null) {
          XMLBuilder ownerBuilder = rootBuilder.elem("Owner");
-         ownerBuilder.elem("ID").text(acl.getOwner().getId()).up();
+         ownerBuilder.elem("ID").text(acl.getOwner().getId());
          if (acl.getOwner().getDisplayName() != null) {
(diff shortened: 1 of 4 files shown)


nacx:
LGTM. Have you been able to run the live tests against aws-s3?

JnRouvignac:
No, unfortunately I do not have an AWS account, so I can't directly test myself :(

nacx:
@gaul any chance you could run the aws-s3 tests for this PR? I think the change is safe, but I'm wondering why this very old code manually copied the ACL objects.

JnRouvignac:
BTW are you interested in removing the dependency to `XMLBuilder`?

This adds a dependency to the `com.jamesmurty.utils:java-xmlbuilder` jar and a transitive dependency to the `net.iharder:base64`  jar (which is superseded by `java.util.Base64` with java 8 anyway). They are 18kb each approximately and not OSGi compatible. They are not huge, but it's less API surface and it's less things to change when trying to use jclouds in an OSGi context (they need to be replaced by OSGi compatible bundles like `org.apache.servicemix.bundles.java-xmlbuilder`).

Of course that would be a separate PR once this PR gets approved and merged.

gaul:
Successfully ran the integration tests against aws-s3.  We are interested in any OSGi improvements, if they remove dependencies or otherwise.  Thank you for your contribution @JnRouvignac!

JnRouvignac:
Excellent! Thank you very much!
I'll write to the mailing list when I'll have a little more time to work on removing these dependencies.

PR #94, opened 2021-01-31
Allow setting BlobAccess in LocalBlobStore.putBlob

gaul:
This addresses a race condition with filesystem users.

### DIFF
--- a/blobstore/src/main/java/org/jclouds/blobstore/LocalStorageStrategy.java
+++ b/blobstore/src/main/java/org/jclouds/blobstore/LocalStorageStrategy.java
@@ -114,14 +114,18 @@ public interface LocalStorageStrategy {
      */
     Blob getBlob(String containerName, String blobName);
 
+    @Deprecated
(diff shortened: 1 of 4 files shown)


nacx (approved):

blobstore/src/main/java/org/jclouds/blobstore/LocalStorageStrategy.java
```diff
@@ -114,14 +114,18 @@
      */
     Blob getBlob(String containerName, String blobName);
 
+    @Deprecated
+    String putBlob(String containerName, Blob blob) throws IOException;
```
jglick, line 118: Should this not have been

```java
default String putBlob(String containerName, Blob blob) throws IOException {
    return putBlob(containerName, blob, BlobAccess.PRIVATE);
}
```

?

gaul: I think it could have been although this would be the first default method in jclouds since 2.3.0 was the first version to require Java 8.

PR #97, opened 2021-03-01
BindACLToXMLPayloadTest.java: ADDED

JnRouvignac:
BindBucketLoggingToXmlPayloadTest.java:
Avoid using the guice injector which triggers an NPE

PR #98, opened 2021-03-01
S3 APIs: Remove the dependency to com.jamesmurty.utils:java-xmlbuilder

JnRouvignac:
XMLBuilder is a very nice fluent API for building XML documents.
However, it is only used by a tiny portion of jclouds: the s3 api. The
use of the XMLBuilder class requires a dependency to
com.jamesmurty.utils:java-xmlbuilder jar and a transitive
dependency to the net.iharder:base64 jar (superseded by
java.util.Base64 in java 8). They are 18kb each approximately and they
not OSGi compatible. They are not huge, but they represent more API
surface and more things to change when trying to use jclouds in an OSGi
context (they need to be replaced by OSGi compatible bundles like
org.apache.servicemix.bundles.java-xmlbuilder).

This commit replaces the use of XMLBuilder by a direct use of the
javax.xml and org.w3c.dom APIs.

I hope retesting will be minimal, and most of this code is covered by
unit tests.

### DIFF
--- a/apis/s3/pom.xml
+++ b/apis/s3/pom.xml
@@ -41,10 +41,6 @@
 
   <dependencies>
     <dependency>
-      <groupId>com.jamesmurty.utils</groupId>
-      <artifactId>java-xmlbuilder</artifactId>
-    </dependency>
-    <dependency>
       <groupId>org.apache.jclouds.api</groupId>
       <artifactId>sts</artifactId>
       <version>${project.version}</version>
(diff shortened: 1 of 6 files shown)


nacx:
LGTM. Thanks! Can you rebase so the build runs including your change in https://github.com/apache/jclouds/pull/97, which introduces the test for the ACL binder?

JnRouvignac:
Of course! Done.

nacx:
Thx!

PR #99, opened 2021-03-08
Upgrade gson from version 2.8.5 to 2.8.6

JnRouvignac:


gurkerl83:
@JnRouvignac We should not do this right now. In https://github.com/apache/jclouds/pull/78 we agreed we pin the version of GSON to 2.8.5. To identify the related parts in the merge request discussion, please look for the 2.8.5 version.

JnRouvignac:
Oops, sorry I did not know about it.
Thanks for the heads-up!

nacx:
@gurkerl83 could you add a small comment in the `pom.xml` next to the gson version explaining the current constraints and why we're pinning it to that version? That could help readers have the context around that version and limitations.

PR #100, opened 2021-03-18
Various code cleanups - mainly removing explicit type arguments

JnRouvignac:


gaul:
Did you make these changes in some automated fashion?  I am not sure how I would review this in any sane way.

I would also caution that these kinds of formatting changes are harmless in one sense but don't deliver any user benefit.  They also have a cost in maintainer/reviewer time that is small and shrinking.

JnRouvignac:
Ok. I wondered if that was too much to take. Apparently it is.
What I can do is submit one PR per commit and make sure to only submit changes that are 100% automated and explain how they were automated.
I will close this PR and submit new ones following what I said above.

PR #101, opened 2021-03-19
Upgrade to error-prone 2.5.1

gaul:
This addresses issues seen with more recent Java and Maven versions.
Enabled by requiring Java 8.  Also suppress several new warnings and
remove stale ones.  This commit also adds the missing jaxb and xml
dependencies needed with newer Java versions.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -54,8 +54,16 @@
       <artifactId>guice-assistedinject</artifactId>
     </dependency>
     <dependency>
+      <groupId>javax.annotation</groupId>
+      <artifactId>javax.annotation-api</artifactId>
+    </dependency>
+    <dependency>
       <groupId>javax.ws.rs</groupId>
       <artifactId>javax.ws.rs-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>javax.xml.bind</groupId>
+      <artifactId>jaxb-api</artifactId>
     </dependency>
     <dependency>
       <groupId>org.osgi</groupId>
--- a/project/pom.xml
+++ b/project/pom.xml
@@ -253,7 +253,7 @@
     <easymock.version>3.4</easymock.version>
 
     <!-- Static analysis dependency versions -->
-    <error_prone.version>2.1.0</error_prone.version>
+    <error_prone.version>2.5.1</error_prone.version>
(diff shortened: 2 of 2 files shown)


gaul:
This need some more investigation.  The intent is to make modern Java and Maven versions work by default since otherwise I need to specify older ones on my desktop.  I believe that error-prone before Java 9 and maybe some earlier Java 8 versions must do strange things to hook the compiler.  Perhaps the easiest approach is to disable error-prone on Java 8 via profiles?

gaul:
Alternatively we can just specify openjdk11 for travis.yml.  Note that this would not change any of the user requirements for jclouds.

gurkerl83:
I am open to using source code quality assurance tools; however, I have a few requirements for integrating and configured them into a project.

The following points relate to the Maven configuration of the project module; this is already very extensive in terms of Maven plugins used and corresponding configuration detail.

Note: I know Maven well enough to say that a great detail of configuration prevents the understanding of essential steps involved in a build process; beginners cannot differentiate crucial from optional build steps. 

An excess of configuration in the individual build steps also makes it difficult to adapt configurations once they have been integrated, or to remove them completely, because a problem that was fundamental at time A has been solved at time B by general adaptations or developments, for example in the programming language or the development tools.

When I made superficial rebuilds to the Maven build a few weeks ago, I also researched the error-prone configuration. This has been in the POM file for a long time and was copy-and-pasted from the Error-Prone project page. This is precisely the kind of situation to avoid; when developers leave the project, the expertise gets lost, and complex configuration fragments remain in a project for a long time.  

Although many of the rules included by the Error-Prone plugin are also supported by development tools such as IntelliJ, there is still a permission base to use the plugin.

To further minimize the project module's configuration, I tend to put the configuration of the plugin or the configuration of the Maven compiler plugin into a settings.xml and call Maven accordingly. The configuration of the compiler options could be passed as a property to the Maven compiler plugin. To support different JDKs, perhaps create additional settings.xml files. 

Due to more JDK variants' support, the profiles will grow; a separation at this level dedicated settings.xml files tailored to one platform would prevent a configuration mess.

In any case, efforts to improve the build setup should be geared towards reducing the build configuration's size.

Thx!

gaul:
Let's revisit this after fixing the Java compatibility issues.

PR #103, opened 2021-04-04
Do not assume Owner ID comes first in XML for S3

timuralp:
Currently, jclouds assumes that the ListBuckets response follows a
specific order: the <Owner> tag, followed by the <Buckets> tag. Within
the <Owner> tag, jclouds further assumes that the <ID> must occur before
the <DisplayName> tag. If the XML body does not adhere to this order,
the parser throws a NullPointerException.

DigitalOcean spaces does not adhere to this order and returns the
<DisplayName> tag before the <ID> tag. The patch changes the parser to
not depend on the order of the tags.

### DIFF
--- a/apis/s3/src/main/java/org/jclouds/s3/xml/ListAllMyBucketsHandler.java
+++ b/apis/s3/src/main/java/org/jclouds/s3/xml/ListAllMyBucketsHandler.java
@@ -39,6 +39,7 @@ public class ListAllMyBucketsHandler extends ParseSax.HandlerWithResult<Set<Buck
 
    private Set<BucketMetadata> buckets = Sets.newLinkedHashSet();
    private CanonicalUser currentOwner;
+   private String currentDisplayName;
    private StringBuilder currentText = new StringBuilder();
 
    private final DateService dateParser;
@@ -48,6 +49,7 @@ public class ListAllMyBucketsHandler extends ParseSax.HandlerWithResult<Set<Buck
    @Inject
    public ListAllMyBucketsHandler(DateService dateParser) {
       this.dateParser = dateParser;
+      this.currentOwner =  new CanonicalUser();
    }
 
    public Set<BucketMetadata> getResult() {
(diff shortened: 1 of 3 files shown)


apis/s3/src/main/java/org/jclouds/s3/xml/ListAllMyBucketsHandler.java
```diff
@@ -56,9 +57,18 @@ public ListAllMyBucketsHandler(DateService dateParser) {
...
+         } else {
+            currentOwner = new CanonicalUser(
+               currentOrNull(currentText), currentDisplayName);
+         }
       } else if (qName.equals("DisplayName")) {
-         currentOwner.setDisplayName(currentOrNull(currentText));
+         if (currentOwner == null) {
+            currentDisplayName = currentOrNull(currentText);
+         } else {
+            currentOwner.setDisplayName(currentOrNull(currentText));
```
gaul, line 70: The original code is wrong here.  Instead what I think would improve it is storing `id` and `displayName` as `String`s then putting this conditional in the Bucket clause.  Note that you will want to set these fields to `null` as well.

gaul (changes requested):

apis/s3/src/main/java/org/jclouds/s3/xml/ListAllMyBucketsHandler.java
timuralp: I see, yes, that's better, although in the current patch and suggested patch there is the latent bug of having the `<Owner>` element appear _after_ the bucket listing. What do you think about adding an empty constructor for the `CanonicalUser` class and setting the values once they are parsed?

timuralp: I updated the patch with a test for a reverse order of <Buckets> and <Owner> tags and expanded the CanonicalUser class to allow setting the ID and a public constructor. This allowed me to change the parser to avoid relying on the order of the elements when processing the XML.

gaul:
Thank you for your contribution @timuralp!

PR #104, opened 2021-04-14
Remove error-prone

gaul:
This does strange things with the java compiler which makes it
incompatible with more modern Java versions.


gaul:
@gurkerl83 jclouds needs some more work with javax.annotation and javax.xml.bind to compile with Java 11 but this is a step in the right direction.

gurkerl83:
@gaul I totally agree with the decision to remove error-prone, it makes things simpler. Thx!

nacx (approved):

PR #106, opened 2021-06-23
JCLOUDS-1580 - Add support for lowercase metadata headers

eebert3:
The issue happens if a cloud provider returns lowercase metadata headers, for example: "x-object-meta-apiversion" instead of "X-Object-Meta-ApiVersion"

In that case, BlobStore.blobMetadata(CONTAINER, PATH).getUserMetadata()
incorrectly returns an empty map.

This happens because the code is looking for the exact String "-Meta-" (case-sensitive).

This checkin allows to handle metadata headers of any case, and also adds a unit test for that situation.


gaul:
Thank you for your contribution @eebert3!

PR #109, opened 2021-07-10
Ignore copyOf method when testing with modern Java

gaul:


gaul:
Curious to see if CI succeeds given the travis-ci.org turndown.  We may need to do some work to move to travis-ci.com.

PR #111, opened 2021-07-10
Add missing javax dependencies for modern Java

gaul:
Java 9 modularization removed these.


gaul:
This is my best guess of what we need to do but it is mostly cobbled together from StackOverflow.  Let's see what CI says about older Java compatibility.

gaul:
Let me try upgrading animal-sniffer separately in #112.

gaul:
Still iterating on this; see #113 and #114.

PR #115, opened 2021-07-15
Remove tests which rely on an isolated ClassLoader

gaul:
Java 9 returns `jdk.internal.loader.ClassLoaders$AppClassLoader` instead
of `URLClassLoader` so we cannot create an isolated `ClassLoader`.


gaul:
Maybe there is a better solution to this but I'm not too familiar with `ClassLoader` issues.

PR #116, opened 2021-07-26
Fix location header case sensitive

pimuzzo:
Fix https://issues.apache.org/jira/browse/JCLOUDS-1569


gaul:
Thank you for your contribution @pimuzzo!  I will fix up the other calls in a subsequent commit.

PR #118, opened 2021-08-04
Store the MPU ETag for the transient blobstore

timuralp:
JCLOUDS-1582: fixes a bug in the transient blobstore where after
uploading a multipart upload, GET/HEAD returns the hash of the content,
rather than the MPU ETag.

### DIFF
--- a/blobstore/src/main/java/org/jclouds/blobstore/TransientStorageStrategy.java
+++ b/blobstore/src/main/java/org/jclouds/blobstore/TransientStorageStrategy.java
@@ -193,7 +193,14 @@ public class TransientStorageStrategy implements LocalStorageStrategy {
          Closeables2.closeQuietly(input);
       }
 
-      Blob newBlob = createUpdatedCopyOfBlobInContainer(containerName, blob, payload, actualHashCode);
+      String eTag = null;
+      if (blob.getMetadata() != null) {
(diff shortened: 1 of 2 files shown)


blobstore/src/main/java/org/jclouds/blobstore/TransientStorageStrategy.java
```diff
@@ -193,7 +195,14 @@ public String putBlob(final String containerName, final Blob blob, BlobAccess ac
          Closeables2.closeQuietly(input);
       }
 
-      Blob newBlob = createUpdatedCopyOfBlobInContainer(containerName, blob, payload, actualHashCode);
+      String eTag = null;
+      if (blob.getMetadata() != null) {
+         eTag = blob.getMetadata().getETag();
+      }
+      if (eTag == null || !MPU_ETAG_FORMAT.matcher(eTag).matches()) {
```
gaul, line 202: Why does this need to sanity check the format?  Is there a danger to allowing an arbitrary input?

timuralp: That's fair. Removed the format check.

gaul:
Thank you for your contribution @timuralp!

PR #127, opened 2021-12-17
JCLOUDS-1590: Promote glacier to core

gaul:


apis/glacier/pom.xml
```diff
@@ -0,0 +1,249 @@
...
+      <id>apache-snapshots</id>
+      <url>https://repository.apache.org/content/repositories/snapshots</url>
+      <releases>
+        <enabled>false</enabled>
+      </releases>
+      <snapshots>
+        <enabled>true</enabled>
+      </snapshots>
+    </pluginRepository>
+  </pluginRepositories>
```
nacx, line 70: We should probably remove this from here and just inherit these settings?

nacx (approved):

apis/glacier/pom.xml
gaul: Done.

nacx (approved):

PR #128, opened 2021-12-24
Update ClientCredentialsClaims.java

SATYANAN-ANAND:
JCLOUDS-1591: OAuth: ClientCredentialsJWTBearerTokenFlow.filter method throws Null Pointer Exception

### DIFF
--- a/apis/oauth/src/main/java/org/jclouds/oauth/v2/domain/ClientCredentialsClaims.java
+++ b/apis/oauth/src/main/java/org/jclouds/oauth/v2/domain/ClientCredentialsClaims.java
@@ -54,7 +54,7 @@ public abstract class ClientCredentialsClaims {
     public abstract String jti();
 
     @SerializedNames({ "iss", "sub", "aud", "exp", "nbf", "jti" })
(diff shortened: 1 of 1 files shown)


apis/oauth/src/main/java/org/jclouds/oauth/v2/domain/ClientCredentialsClaims.java
```diff
@@ -49,6 +50,7 @@
     public abstract long nbf();
 
     /** "JWT ID", a unique identifier for the JWT. */
+    @Nullable
```
gaul, line 53: Should this also annotate the `create` method with `@Nullable` as the [AutoValue documentation suggests](https://github.com/google/auto/blob/master/value/userguide/howto.md#nullable)?

SATYANAN-ANAND: Modified as per the comment and commit the changes.

gaul:
Thank you for your contribution @SATYANAN-ANAND!

PR #129, opened 2022-01-10
JCLOUDS-1594: Allow overriding S3 signer

gaul:
Previously s3 always used v2 and aws-s3 always used v4.  Now s3
defaults to v2 and can override to v4.  Note that this does not change
BlobRequestSigner.


nacx (approved):
LGTM! Thx!

gaul:
@jiffysrc please test.

PR #130, opened 2022-01-13
JCLOUDS-1587 - S3 Blobstore: Support Bucket Configuration Options - Versioning, Lifecycle, Encryption

jiffysrc:
Changes included in - https://issues.apache.org/jira/browse/JCLOUDS-1587 after removing noise changes.

Support following features in jclouds s3 blobstore

Configure the bucket for versioning and get the versioning state of a bucket
Configure the bucket for encryption and get the encryption state of a bucket
Configure the bucket for lifecycle rules and get the lifecycle rules of a bucket
List different versions of an Object

### DIFF
--- a/apis/s3/src/main/java/org/jclouds/s3/blobstore/S3BlobStore.java
+++ b/apis/s3/src/main/java/org/jclouds/s3/blobstore/S3BlobStore.java
@@ -29,13 +29,6 @@ import javax.inject.Inject;
 import javax.inject.Provider;
 import javax.inject.Singleton;
 
-import com.google.common.base.Function;
-import com.google.common.base.Predicate;
-import com.google.common.base.Supplier;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.Iterables;
-
 import org.jclouds.blobstore.BlobStoreContext;
 import org.jclouds.blobstore.ContainerNotFoundException;
 import org.jclouds.blobstore.domain.Blob;
@@ -81,6 +74,13 @@ import org.jclouds.s3.options.PutBucketOptions;
 import org.jclouds.s3.options.PutObjectOptions;
 import org.jclouds.s3.util.S3Utils;
 
+import com.google.common.base.Function;
+import com.google.common.base.Predicate;
+import com.google.common.base.Supplier;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
+
 @Singleton
 public class S3BlobStore extends BaseBlobStore {
    private final S3Client sync;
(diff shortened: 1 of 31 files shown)


blobstore/src/main/java/org/jclouds/blobstore/domain/internal/MutableStorageMetadataImpl.java
```diff
@@ -38,6 +38,8 @@
    private Date lastModified;
    private Long size;
    private Tier tier;
+   private String versionId;
+   private String isLatest;
```
gaul, line 42: Should this be a `boolean`?

apis/s3/src/test/java/org/jclouds/s3/S3ClientTest.java
```diff
@@ -501,6 +504,86 @@ public void testDisableBucketLogging() throws SecurityException, NoSuchMethodExc
...
+      assertRequestLineEquals(request, "PUT https://bucket." + url + "/?lifecycle=true HTTP/1.1");
+      assertNonPayloadHeadersEqual(request, "Host: bucket." + url + "\n");
+      assertPayloadEquals(request, "hello", "text/plain",
+              false);
+
+      assertResponseParserClassEquals(method, request, BucketConfigurationHandler.class);
+      assertSaxResponseParserClassEquals(method, null);
+      assertFallbackClassEquals(method, null);
+
+      checkFilters(request);
```
gaul, line 583: Can you add a live test for adding multiple objects with the same name and different versions?  These unit tests are good but comparing against live AWS will give me more confidence about this PR.

apis/s3/src/main/java/org/jclouds/s3/blobstore/S3BlobStore.java
```diff
@@ -133,6 +134,18 @@ public boolean containerExists(String container) {
       return sync.bucketExists(container);
    }
 
+   public String getBucketConfiguration(String bucketName, BucketConfigOptions httpOptions ) {
+      return sync.getBucketConfiguration(bucketName, httpOptions);
+   }
+
+   public String deleteBucketConfiguration(String bucketName, BucketConfigOptions httpOptions ) {
```
gaul, line 141: Extra whitespace here and above.

apis/s3/src/main/java/org/jclouds/s3/options/ListBucketOptions.java
```diff
@@ -96,6 +96,16 @@ public String getDelimiter() {
       return getFirstQueryOrNull("delimiter");
    }
 
+   public ListBucketOptions versions(Boolean versions) {
```
gaul, line 99: Should this be a `boolean` if it cannot be `null`?

blobstore/src/main/java/org/jclouds/blobstore/domain/internal/StorageMetadataImpl.java
```diff
@@ -40,6 +40,10 @@
    @Nullable
    private final String eTag;
    @Nullable
+   private final String versionId;
+   @Nullable
+   private final String isLatest;
```
gaul, line 45: `boolean`?  Need to propagate this throughout.

blobstore/src/main/java/org/jclouds/blobstore/domain/internal/StorageMetadataImpl.java
```diff
@@ -60,6 +64,23 @@ public StorageMetadataImpl(StorageType type, @Nullable String id, @Nullable Stri
...
       this.tier = tier;
+      this.versionId = null;
+      this.isLatest = null;
+   }
+
+   public StorageMetadataImpl(StorageType type, @Nullable String id, @Nullable String name,
+         @Nullable Location location, @Nullable URI uri, @Nullable String eTag,
+         @Nullable Date creationDate, @Nullable Date lastModified,
+         Map<String, String> userMetadata, @Nullable Long size, Tier tier, @Nullable String versionId) {
+      super(id, name, location, uri, userMetadata);
```
gaul, line 75: Please make the constructor with all parameters the only one that sets fields and the existing one with fewer parameters can call `this(...)` with a default value for `versionId`.

blobstore/src/main/java/org/jclouds/blobstore/options/ListContainerOptions.java
```diff
@@ -58,6 +59,12 @@ public ListContainerOptions() {
       this.delimiter = delimiter;
    }
 
+   ListContainerOptions(Integer maxKeys, String marker, String dir, boolean recursive,
+            boolean detailed, String prefix, String delimiter, boolean versions) {
+      this(maxKeys, marker, dir, recursive, detailed, prefix, delimiter);
+      this.versions = versions;
```
gaul, line 65: Same comment about making this the primary method and add a `@Deprecated` annotation to the old method.

gaul (changes requested):
Sorry about my delayed response -- this is my first interaction with object versioning but something that interests me.  I am primarily concerned about the lack of live tests.  You can run these via:

```
mvn integration-test -pl :aws-s3 -Plive \
        -Dtest.aws-s3.identity="${JCLOUDS_IDENTITY}" \
        -Dtest.aws-s3.credential="${JCLOUDS_CREDENTIAL}" \
```

Outside the scope of this PR but I am interested in adding support for versions to S3Proxy and the jclouds portable abstraction.  Google Cloud Storage already has support but we probably need Azure support for this to make sure we make this portable enough.

gaul:
@jiffysrc can we make progress on this PR?  I would like to release jclouds 2.5.0 next month and include this functionality.

gaul:
@jiffysrc do we have a path forward on this PR?

gaul:
@jiffysrc can we move forward with this or should I close it?

gaul:
Please re-open if this is still relevant.

PR #131, opened 2022-01-22
JCLOUDS-1595: Add URL encoding to string vals in ListBucketOptions.

JimSermersheim:
https://issues.apache.org/jira/browse/JCLOUDS-1595


JimSermersheim:
Declined after discovering a case where values were already url encoded. The simplistic fix causes these values to be double-encoded. I'll probably submit again where the encode step decodes and re-encodes.

gaul:
Thanks for checking the double-quoting.  Note that we only merge PRs against the master branch and do not plan further 2.4.x releases.

PR #142, opened 2022-05-02
Adding access key auth support for openstack V3

DImuthuUpe:
Please refer to the discussion thread [1]

[1] https://lists.apache.org/thread/2p7qc20rn19r3yhsfd63p3o3h587f7oh

### DIFF
--- a/apis/openstack-keystone/src/main/java/org/jclouds/openstack/keystone/v3/auth/V3AuthenticationApi.java
+++ b/apis/openstack-keystone/src/main/java/org/jclouds/openstack/keystone/v3/auth/V3AuthenticationApi.java
@@ -27,6 +27,9 @@ import org.jclouds.openstack.keystone.auth.AuthenticationApi;
 import org.jclouds.openstack.keystone.auth.domain.PasswordCredentials;
 import org.jclouds.openstack.keystone.auth.domain.TenantOrDomainAndCredentials;
 import org.jclouds.openstack.keystone.auth.domain.TokenCredentials;
+import org.jclouds.openstack.keystone.auth.domain.ApiAccessKeyCredentials;
+import org.jclouds.openstack.keystone.auth.domain.AuthInfo;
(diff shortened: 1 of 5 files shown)


smarru:
@gaul drawing your attention, can you please review and provide us feedback, we would like to use this in Apache Airavata

apis/openstack-keystone/src/main/java/org/jclouds/openstack/keystone/v3/auth/V3AuthenticationApi.java
```diff
@@ -57,4 +60,10 @@ public interface V3AuthenticationApi extends AuthenticationApi, Closeable {
    @Override
    Token authenticateToken(TenantOrDomainAndCredentials<TokenCredentials> credentials);
 
+   @Named("token:create")
+   @POST
+   @ResponseParser(ParseTokenFromHttpResponse.class)
+   @MapBinder(BindAccessKeyAuthToJsonPayload.class)
+   @Override
+   AuthInfo authenticateAccessKey(TenantOrDomainAndCredentials<ApiAccessKeyCredentials> credentials);
```
nacx, line 68: Please add the corresponding unit tests here: https://github.com/apache/jclouds/blob/master/apis/openstack-keystone/src/test/java/org/jclouds/openstack/keystone/v3/auth/V3AuthenticationApiMockTest.java

And the live test as well here: 
https://github.com/apache/jclouds/blob/master/apis/openstack-keystone/src/test/java/org/jclouds/openstack/keystone/v3/auth/V3AuthenticationApiLiveTest.java

And include in a comment in the PR the output of running the live tests.

nacx (commented):
Thanks!

gaul (changes requested):
@DImuthuUpe Can you add tests?

gaul:
@DImuthuUpe Can we move forward with this or should I close it?

gaul:
Please re-open if this is still relevant.

edipdemirbilek:
@gaul I have run into the same issue. i wonder is this was ever fixed.

PR #145, opened 2022-06-12
JCLOUDS-1609: Upgrade to BouncyCastle 1.71

gaul:
Release notes:

https://www.bouncycastle.org/releasenotes.html#r1rv71


gaul:
@coheigea I don't use this driver so it would be nice if someone from the community could test it.

PR #146, opened 2022-07-24
Synchronize ByteSourcePayload in LocalBlobStore

gaul:
`ByteSourcePayload's.openStream` is not thread safe and lack of
synchronization can throw `ArrayIndexOutOfBoundsExceptions`.
Fixes gaul/s3proxy#303.

### DIFF
--- a/blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
+++ b/blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
@@ -564,7 +564,9 @@ public final class LocalBlobStore implements BlobStore {
 
       InputStream is = null;
       try {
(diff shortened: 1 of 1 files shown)


gaul:
I think there is a narrower fix where `LocalBlobStore` could synchronize its calls sites instead.  @nacx what do you think?

nacx (approved):
LGTM!

nacx:
If the LocalBlobStore can sync ina. narrower scope that would be much better

gaul:
> If the LocalBlobStore can sync ina. narrower scope that would be much better

Done.

gaul:
Something seems strange here -- every call to `getBlob` should return a new `Payload`...

nacx (approved):

gaul:
Superseded by #150.

PR #153, opened 2022-08-09
JCLOUDS-1617: Fix HTTPS support in OkHttpCommandExecutorService

SATYANAN-ANAND:
Added support for  proxy server type = HTTPS

### DIFF
--- a/drivers/okhttp/src/main/java/org/jclouds/http/okhttp/DelegatingSocketFactory.java
+++ b/drivers/okhttp/src/main/java/org/jclouds/http/okhttp/DelegatingSocketFactory.java
@@ -1,0 +1,66 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


drivers/okhttp/src/main/java/org/jclouds/http/okhttp/DelegatingSocketFactory.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+package org.jclouds.http.okhttp;
+
+import java.io.IOException;
+import java.net.InetAddress;
+import java.net.Socket;
+
+import javax.net.SocketFactory;
+
+public class DelegatingSocketFactory extends SocketFactory {
```
nacx, line 25: Why is this class needed?

SATYANAN-ANAND: Hi 

I was referring below issue to fix this issue. 

https://github.com/square/okhttp/issues/6561#issuecomment-782308646
https://github.com/square/okhttp/issues/6561#issuecomment-782292563

nacx: My point is that unless I'm missing something this `DelegatingSocketFactory` doesn't add any logic and it's doing nothing. Why do we need it?

SATYANAN-ANAND: Hi,

Builder.setSSLSocketFactory is deprecated which force to use Builder.setSocketFactory
but setSocketFactory doesnot accept SSLSocketfactory.getDefault() and throws "java.lang.IllegalArgumentException: socketFactory instanceof SSLSocketFactory" at **runtime**

To overcome the above problem, a delegator class called "DelegatingSocketFactory" created which delegates instance of SSLSocketFactory to SocketFactory

nacx: Thanks for the explanation!
Can you add this motivation in a comment in the code so anyone reading it can easily get why it is done this way?

nacx (approved):

PR #155, opened 2022-09-15
Jclouds 1615

ChJung1:
Hi @gaul,
This is my first github pull request, so forgive me if i did something the wrong way.
I tried to add a test, but feel that it kind of doesn't test the core of the problem. 
Feedback welcome.
    ChJung1


providers/azureblob/src/test/java/org/jclouds/azureblob/blobstore/AzureBlobStoreTest.java
```diff
@@ -0,0 +1,57 @@
...
+       checkBlockIdForPartNumber(868096);
+       checkBlockIdForPartNumber(-1);
+       checkBlockIdForPartNumber(-1023);
+   }
+   
+   private void checkBlockIdForPartNumber(int partNumber) {
+       String blockId = AzureBlobStore.makeBlockId(partNumber);
+       // if (!blockId.matches(VALIDATION_REGEX)){
+       //    System.out.println("--------------------------------------------- " + partNumber + " -> '" + blockId + "' -> " + blockId.matches(VALIDATION_REGEX));
+       //}
```
gaul, line 54: Please remove this commented code.

providers/azureblob/src/test/java/org/jclouds/azureblob/blobstore/AzureBlobStoreTest.java
```diff
@@ -0,0 +1,57 @@
...
+import org.testng.annotations.Test;
+
+/**
+ * Tests behavior of {@code AzureBlobStore}
+ */
+// NOTE:without testName, this will not call @Before* and fail w/NPE during surefire
+@Test(groups = "unit", testName = "AzureBlobStore")
+public class AzureBlobStoreTest {
+    
+    static String VALIDATION_REGEX = "[a-zA-Z0-9\\-_=]*";
```
gaul, line 30: Can you make this `private` and `final` and use `Pattern.compile?`?

gaul (changes requested):
Please squash commits.

providers/azureblob/src/test/java/org/jclouds/azureblob/blobstore/AzureBlobStoreTest.java
ChJung1: ... removed

providers/azureblob/src/test/java/org/jclouds/azureblob/blobstore/AzureBlobStoreTest.java
ChJung1: ... i did so ... hope you like it.

gaul:
Thank you for your contribution @ChJung1!

inno-steffg:
I have tested some simple things with the current JCloud master branch code base against an Azure storage account and run into an error message for a 4 gigabyte upload, split into 5 MByte chunks in a multipart upload:
2023-07-25_11h40_04

Could it be, that the base64 encoding in the url version isn't accepted by Azure?

ChJung1:
Hi Steffen Gürtler,
I did the pull request for the encoding with Base64Url, because with Base64 i got an error similar to yours.

Base64Url is meant to ensure you can use the string in urls.

The string you have contains a minus sign, so this is indeed Base64Url, and not Base64, so my fix is in there.

Your problem looks very similar to mine, so is it maybe the "==" this time?

 

BTW, while changing the code i wondered if we could implement makeBlockId() by using String.valueOf() instead,

which is much simpler code, a shorter string, and uses only digits (as Block-ID is an integer and probably positive)

Feel free to try that, and if it helps you can maybe also remove that not-so-helpful test I added.

 

Andrew Gaul will then hopefully approve your pull request.

 

Good luck!

    ChJung1

 

Gesendet: Dienstag, 25. Juli 2023 um 18:07 Uhr
Von: "Steffen Gürtler" ***@***.***>
An: "apache/jclouds" ***@***.***>
Cc: "ChJung1" ***@***.***>, "Mention" ***@***.***>
Betreff: Re: [apache/jclouds] Jclouds 1615 (PR #155)


 

I have tested some simple things with the current JCloud master branch code base against an Azure storage account and run into an error message for a 4 gigabyte upload, split into 5 MByte chunks in a multipart upload:


Could it be, that the base64 encoding in the url version isn't accepted by Azure?

gaul:
Please open a Jira issue for this regression.

PR #156, opened 2022-11-11
activate Reproducible Builds

hboutemy:
see https://maven.apache.org/guides/mini/guide-reproducible-builds.html


gaul:
Thank you for your contribution @hboutemy!

PR #158, opened 2022-11-26
Add ByteSource wrapper for ByteBuffer

gaul:


gaul:
Proposed in response to: https://stackoverflow.com/questions/74398323/how-to-upload-direct-buffer-gotten-from-jni-to-s3-directly/74580865

PR #161, opened 2022-12-26
JCLOUDS-1371: JCLOUDS-1488: optimize fs prefix

gaul:
This reduces the number of stat calls required when prefix is deep in the filesystem hierarchy.  Further optimizations to delimiter are possible. References gaul/s3proxy#473.


gaul:
@timuralp maybe interesting to you

PR #162, opened 2023-01-21
Atomically replace objects in filesystem putBlob

gaul:
Fixes gaul/s3proxy#490.  This fixes a regression from 41ce90ec360a46cb7b03f7cb8b66e01a113aa5b6.  Continue to support Windows logic although it is unclear if this is necessary.


gaul:
I cannot test this on Windows so I don't know if it actually helps there.

PR #163, opened 2023-01-22
Lazily open parts during `LocalBlobStore` complete MPU

gaul:
This removes a previous workaround for opening too many `FileInputStream` and exhausting rlimits.

### DIFF
--- a/blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
+++ b/blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
@@ -28,7 +28,6 @@ import static com.google.common.collect.Sets.newTreeSet;
 import static org.jclouds.blobstore.options.ListContainerOptions.Builder.recursive;
 
 import java.io.File;
-import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
@@ -739,14 +738,8 @@ public final class LocalBlobStore implements BlobStore {
       Payload payload;
       try {
          InputStream is = blob.getPayload().openStream();
-         if (is instanceof FileInputStream) {
-            // except for FileInputStream since large MPU can open too many fds
-            is.close();
-            payload = blob.getPayload();
-         } else {
-            blob.resetPayload(/*release=*/ false);
-            payload = new InputStreamPayload(is);
-         }
+         blob.resetPayload(/*release=*/ false);
+         payload = new InputStreamPayload(is);
       } catch (IOException ioe) {
          throw new RuntimeException(ioe);
       }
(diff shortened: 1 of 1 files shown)


blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
```diff
@@ -739,14 +738,8 @@ public Blob getBlob(String containerName, String key, GetOptions options) {
       Payload payload;
       try {
          InputStream is = blob.getPayload().openStream();
-         if (is instanceof FileInputStream) {
-            // except for FileInputStream since large MPU can open too many fds
-            is.close();
```
gaul, line 744: A subsequent PR will return a `FileInputStream` instead of a `ByteSource` to fix a race condition in the filesystem blobstore so this code cannot assume that the `Payload` is a repeatable `ByteSource` that can be re-opened.

jixinchi:
Hi! I'm using s3proxy built on the latest jclouds. The provider configuration is "filesystem", and basedir configuration is a directory where a NAS is mounted.

When I performed a multipart upload, I found that files such as ".nfs00000000100000bf00000005" remained in the directory after uploading. Each file corresponded to one part and the files could not be deleted. The error "Device Busy" was reported. These files will disappear automatically after restarting s3proxy.

I suspected that the file handle may not released, and eventually I found this PR. I restored this workaround code about FileInputStream, then my problem was solved. So I would like to ask why this workaround was deleted. Is there a better solution to solve my problem?

Thank you very much!

gaul:
Eagerly opening all the `FileInputStream` exhausted system limits for large MPU, e.g., 10,000 parts.  If there is a leak of `FileInputStream` then it should be fixed instead of reverting this PR.

jixinchi:
Sorry, I'm not very familiar with jclouds code and can't find a proper way to fix the problem. I tried to put a finally part here to close `is`, but the test failed.

Could you please take a look at this problem and fix it in the future? Thank you very much!

gaul:
Please open a JIRA issue for this although you are more likely to fix this than me judging by my backlog.

jixinchi:
I attempted to apply a JIRA account a few days ago, but haven't received a response yet. It would be greatly appreciated if you could help me record this issue!

I'm trying to fix the bug, but the codes about blob/payload/stream are too complex for me, and it looks like you have some TODOs and ongoing PRs on it. I am not able to fix the problem. What I can do is just to revert the following snippet and wait for your help.
```java
         InputStream is = blob.getPayload().openStream();
         if (is instanceof FileInputStream) {
            // except for FileInputStream since large MPU can open too many fds
            is.close();
            payload = blob.getPayload();
         } else {
            blob.resetPayload(/*release=*/ false);
            payload = new InputStreamPayload(is);
         }
```

PR #164, opened 2023-01-22
JCLOUDS-1371: Optimize filesystem delimiter

gaul:
`populateBlobKeysInContainer` will no longer recurse when the delimiter matches "/".  This makes listing deep hierarchies with a delimiter faster.  Note that the general `LocalBlobStore` handling is still required for the general cases.  This requires removing a bogus test case.  References gaul/s3proxy#473.


gaul:
@timuralp Maybe interesting to you.

PR #165, opened 2023-01-29
Use the same file handle in file system `getBlob`

gaul:
Previously `getBlob` lazily returned a `ByteSource` which could return a different content length than the blob metadata.  Instead eagerly open a `FileInputStream` for consistency.  Note that this has some consequences for `LocalBlobStore.getBlob` when used with ranges.


gaul:
I am still thinking about the range issue.

gaul:
Abandoning this since S3Proxy's filesystem-nio2 provider does what I need: https://github.com/gaul/s3proxy/issues/697.

PR #176, opened 2023-06-06
JCLOUDS-1618: Upgrade to gson 2.10.1

gaul:
~~This also upgrades bnd to 7.0.0-SNAPSHOT to work around a `META-INF/versions/9/module-info.class=module-info` issue.  Obviously we cannot merge this due to the use of a SNAPSHOT version but maybe someone has a better suggestion how to approach this?~~

Removed bnd in #177.

### DIFF
--- a/project/pom.xml
+++ b/project/pom.xml
@@ -221,7 +221,7 @@
     <maven-surefire-plugin.version>3.1.0</maven-surefire-plugin.version>
 
     <!-- General dependency versions -->
-    <gson.version>2.8.9</gson.version>
+    <gson.version>2.10.1</gson.version>
     <guava.version>32.0.0-jre</guava.version>
     <guice.version>5.1.0</guice.version>
 
--- a/core/src/main/java/org/jclouds/json/config/GsonModule.java
+++ b/core/src/main/java/org/jclouds/json/config/GsonModule.java
@@ -38,6 +38,7 @@ import org.jclouds.domain.JsonBall;
 import org.jclouds.domain.LoginCredentials;
 import org.jclouds.json.Json;
 import org.jclouds.json.SerializedNames;
+import com.google.gson.ReflectionAccessFilter;
 import com.google.gson.internal.ConstructorConstructor;
 import com.google.gson.internal.Excluder;
 import com.google.gson.internal.JsonReaderInternalAccess;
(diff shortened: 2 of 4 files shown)


gaul:
CI reports:

> org.codehaus.plexus.component.repository.exception.ComponentLookupException: aQute/bnd/maven/plugin/BndMavenPlugin has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 52.0

I don't know if Java 17 is an intended requirement but I am flagging this.

project/pom.xml
```diff
@@ -217,12 +227,12 @@
     <project.build.outputTimestamp>1668154596</project.build.outputTimestamp>
 
     <!-- Plugin versions -->
-    <bnd.version>5.2.0</bnd.version>
+    <bnd.version>7.0.0-SNAPSHOT</bnd.version>
```
gaul, line 230: I don't understand why we are using bnd in the first place -- is this some kind of OSGi thing?  @jbonofre

gaul: Alternatively we could remove all uses of bnd and `<manifestFile>` from maven-jar-plugin.

gurkerl83:
I have to say it but I guess this completely breaks OSGi compatibility. I introduced the bnd plugin here which really simplified the OSGi support https://github.com/apache/jclouds/commit/8ac994c2b5d509ca2877efdd77c35881d58ac3e3 . Please revert the commit and I will take a look.

gurkerl83:
> CI reports:
> 
> > org.codehaus.plexus.component.repository.exception.ComponentLookupException: aQute/bnd/maven/plugin/BndMavenPlugin has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 52.0
> 
> I don't know if Java 17 is an intended requirement but I am flagging this.

Looking at the error and comparing the bnd-plugin metadata I found that the version in use 5.2 can be updated until 6.0.0.
As you can the version 6.1.0 changes the required jdk version to 17. 

```
5.2
Build-Jdk-Spec: 1.8

5.3
Build-Jdk-Spec: 1.8

6.0.0
Build-Jdk-Spec: 1.8

6.1.0
Build-Jdk-Spec: 17
```

See https://javaalmanac.io/bytecode/versions/

gaul:
No reversion, see https://github.com/apache/jclouds/pull/177#issuecomment-1622645012.

jonesbusy:
Hi,

Any plan to have this released ?

Outdated gson cause now some issues on a Jenkins plugin (artifact-manager-s3-plugin) (https://issues.jenkins.io/browse/JENKINS-72441, https://github.com/jenkinsci/gson-api-plugin/issues/7)

```
[INFO] +- org.apache.jclouds.provider:aws-s3:jar:2.5.0:compile
[INFO] |  \- org.apache.jclouds.api:s3:jar:2.5.0:compile
[INFO] |     +- org.apache.jclouds.api:sts:jar:2.5.0:compile
[INFO] |     |  \- org.apache.jclouds:jclouds-core:jar:2.5.0:compile
[INFO] |     |     +- com.google.code.gson:gson:jar:2.8.9:compile
```

Thanks!

PR #177, opened 2023-06-12
Remove bnd

gaul:
This blocks upgrading gson.  References #176.

### DIFF
--- a/project/pom.xml
+++ b/project/pom.xml
@@ -217,7 +217,6 @@
     <project.build.outputTimestamp>1668154596</project.build.outputTimestamp>
 
     <!-- Plugin versions -->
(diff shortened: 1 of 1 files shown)


gaul:
@jbonofre I guess this is OSGi related but I lack the background to understand its use.  Can we safely remove this without breaking OSGi support?  If not, I think breaking OSGi to move forward with 2.6.0 is a good tradeoff.

gaul:
Sorry no reversions.  We have been talking about this problem for almost a year so the burden is on the community to find a better solution.  If you want to submit a fix that preserves gson 2.10.1 you can try to do this but I plan to cut the next (final) release in a few weeks so you should start now.

PR #185, opened 2023-10-12
check file exists in getContainerAccess

jixinchi:
I found that getBlobAccess checks file exists, but getContainerAccess doesn't check. This causes s3proxy throws exceptions when OPTIONS a non-exist bucket. https://github.com/gaul/s3proxy/pull/561


nacx (approved):
Thanks!

PR #186, opened 2023-10-19
implement support for SharedKey signature

larshagencognite:
This is the recommended signature scheme for Azure, and the only scheme that is supported by the Azurite emulator.

https://learn.microsoft.com/en-us/rest/api/storageservices/authorize-with-shared-key

### DIFF
--- a/providers/azureblob/src/main/java/org/jclouds/azure/storage/filters/SharedKeyLiteAuthentication.java
+++ b/providers/azureblob/src/main/java/org/jclouds/azure/storage/filters/SharedKeyLiteAuthentication.java
@@ -34,6 +34,14 @@ import javax.inject.Named;
 import javax.inject.Provider;
 import javax.inject.Singleton;
 
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Maps;
+import com.google.common.collect.Multimap;
+import com.google.common.collect.Multimaps;
+import com.google.common.collect.Multiset;
+import com.google.common.collect.TreeMultiset;
 import org.jclouds.Constants;
 import org.jclouds.azure.storage.config.AuthType;
 import org.jclouds.azure.storage.util.storageurl.StorageUrlSupplier;
@@ -56,13 +64,7 @@ import com.google.common.base.Function;
 import com.google.common.base.Joiner;
 import com.google.common.base.Strings;
 import com.google.common.base.Supplier;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableMap.Builder;
-import com.google.common.collect.Iterables;
-import com.google.common.collect.Maps;
-import com.google.common.collect.Multimap;
-import com.google.common.collect.Multimaps;
 import com.google.common.io.ByteProcessor;
 import com.google.common.net.HttpHeaders;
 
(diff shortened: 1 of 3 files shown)


larshagencognite:
Not sure how this feature should be introduced, but I think it would be nice to have it as an opt-in feature to begin with. There is a lot of parallel code in `SharedKeyLiteAuthentication.java` now, and it would probably be better to put the logic in two separate classes, but not sure how that would work with the annotations.

Happy to take feedback on the direction.

larshagencognite:
I have validated that I am able to sign in to Azurite using this new signing method, which was not possible with SharedKeyLite.

providers/azureblob/src/main/java/org/jclouds/azure/storage/filters/SharedKeyLiteAuthentication.java
```diff
@@ -187,7 +209,22 @@ public String[] cutUri(URI uri) throws IllegalArgumentException {
...
+      StringBuilder buffer = new StringBuilder();
+      // re-sign the request
+      appendMethod(request, buffer);
+      appendPayloadMetadataForSharedKey(request, buffer);
+      appendHttpHeadersForSharedKey(request, buffer);
+      appendCanonicalizedHeaders(request, buffer);
+      appendCanonicalizedResourceForSharedKey(request, buffer);
+      if (signatureWire.enabled())
+         signatureWire.output(buffer.toString());
+      System.out.println(buffer);
```
nacx, line 225:
```suggestion
```

providers/azureblob/src/main/java/org/jclouds/azure/storage/filters/SharedKeyLiteAuthentication.java
```diff
@@ -203,6 +240,24 @@ public String createStringToSign(HttpRequest request) {
       return buffer.toString();
    }
 
+   private void appendPayloadMetadataForSharedKey(HttpRequest request, StringBuilder buffer) {
```
nacx, line 243: Probably we can just early return if the request payload is null, and don't check it every time?

providers/azureblob/src/main/java/org/jclouds/azure/storage/filters/SharedKeyLiteAuthentication.java
```diff
@@ -216,6 +271,7 @@ public String calculateSignature(String toSign) throws HttpException {
       String signature = signString(toSign);
       if (signatureWire.enabled())
          signatureWire.input(Strings2.toInputStream(signature));
+      System.out.println(signature);
```
nacx, line 274: Change this for logs if you really need this output, otherwise let's just remove this.

providers/azureblob/src/main/java/org/jclouds/azure/storage/filters/SharedKeyLiteAuthentication.java
```diff
@@ -268,6 +329,51 @@ void appendCanonicalizedResource(HttpRequest request, StringBuilder toSign) {
...
+      toSign.append(request.getEndpoint().getRawPath());
+      appendQueryParametersForSharedKey(request, toSign);
+   }
+
+   void appendQueryParametersForSharedKey(HttpRequest request, StringBuilder toSign) {
+      // 3. Append each query parameter as a new line
+      Map<String, Multiset<String>> sortedParams = Maps.newTreeMap();
+      if (request.getEndpoint().getQuery() != null) {
+         String[] params = request.getEndpoint().getQuery().split("&");
+         for (String param : params) {
```
nacx, line 346: Can you use something like?
```java
Map<String, String> params = Splitter.on('&')
   .trimResults()
   .omitEmptyStrings()
   .withKeyValueSeparator('=')
   .split(request.getEndpoint().getQuery());
```

nacx (commented):
Thanks! Overall LGTM.

providers/azureblob/src/main/java/org/jclouds/azure/storage/filters/SharedKeyLiteAuthentication.java
larshagencognite: My bad, this was a debugging aid that slipped into the PR

providers/azureblob/src/main/java/org/jclouds/azure/storage/filters/SharedKeyLiteAuthentication.java
larshagencognite: Yes, this was just a debugging aid, I forgot to remove it.

providers/azureblob/src/main/java/org/jclouds/azure/storage/filters/SharedKeyLiteAuthentication.java
larshagencognite: done

PR #188, opened 2023-11-13
do not remove md in DelimeterFilter then add it by commonPrefixes; just keep it

jixinchi:
If the metadata ends with delimiter, it is removed in `DelimterFilter.apply`, then add it back by loop `commonPrefixes`. But many information in the metadata lost, such as lastModified time. My solution is keeping them in the DelimeterFilter.

### DIFF
--- a/blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
+++ b/blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
@@ -364,6 +364,9 @@ public final class LocalBlobStore implements BlobStore {
             o = prefix + o;
          }
          md.setName(o + delimiter);
+         if (!contents.contains(md)) {
+            contents.add(md);
+         }
          contents.add(md);
       }
       return contents;
@@ -471,7 +474,7 @@ public final class LocalBlobStore implements BlobStore {
       public boolean apply(StorageMetadata metadata) {
          String name = metadata.getName();
          if (prefix == null || prefix.isEmpty()) {
-            return name.indexOf(delimiter) == -1;
+            return name.indexOf(delimiter) == -1 || name.indexOf(delimiter) == name.length() - delimiter.length();
          }
          if (name.startsWith(prefix)) {
             String unprefixedName = name.substring(prefix.length());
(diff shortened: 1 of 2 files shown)


jixinchi:
This PR is related to https://github.com/gaul/s3proxy/pull/574

blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
```diff
@@ -364,6 +364,9 @@ private SortedSet<StorageMetadata> extractCommonPrefixes(SortedSet<StorageMetada
             o = prefix + o;
          }
          md.setName(o + delimiter);
+         if (!contents.contains(md)) {
+            contents.add(md);
+         }
          contents.add(md);
```
gaul, line 370: Doesn't this add `md` twice?

gaul (changes requested):
Can you exercise this change with some test?

blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
jixinchi: This was a typo and I removed it. Thanks!

jixinchi:
> Can you exercise this change with some test?

I add a test `testListBlobEndsWithDelimiterAndDelimiterFilter`

blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
```diff
@@ -364,7 +364,9 @@ private SortedSet<StorageMetadata> extractCommonPrefixes(SortedSet<StorageMetada
             o = prefix + o;
          }
          md.setName(o + delimiter);
-         contents.add(md);
+         if (!contents.contains(md)) {
+            contents.add(md);
+         }
```
gaul, line 369: `contents` is a `Set` so this is the same as before.  Could you revert it?

gaul (changes requested):

blobstore/src/main/java/org/jclouds/blobstore/config/LocalBlobStore.java
jixinchi: It is reverted now. I thought that item will be replaced if I `add` an existing item in `Set`, but actually it will be ignored.

gaul:
Thank you for your contribution @jixinchi!

PR #189, opened 2023-11-22
Parse out of order ListBucket response elements

mlanoe:
Currently, jclouds assumes that the "ID" must occur before the "DisplayName" tag. If the XML body does not adhere to this order, the parser throws a NullPointerException.

Scaleway is a cloud provider that does not adhere to this order and returns the "DisplayName" tag before the "ID" tag. The patch changes the parser to not depend on the order of the tags.

See commit 04feb8f8482e48b6c26166edb3d6fd5a607d858b for a similar fix


gaul:
Thank you for your contribution @mlanoe!

PR #190, opened 2024-01-30
fixes bug #1519

davidsenk:
Fixes the authorization error with b2 application keys.

Tested with s3proxy and curl

https://issues.apache.org/jira/projects/JCLOUDS/issues/JCLOUDS-1519

https://github.com/gaul/s3proxy/issues/298


gaul:
@davidsenk Please investigate CI failures.

davidsenk:
@gaul resolved, https://github.com/davidsenk/jclouds/actions/runs/7953299249 

there were a few bad assumptions in the test requests. YOUR_ACCOUNT_ID is returned in authorize_account_response.json, which is then not used in a few of the *_request.json payloads.

Additionally, the mockwebserver does not get the content-type headers set throughout the test. The "easiest" solution is to fix it so that the mock assert is only looking for application/unknown (default in mockhttpserver).

gaul:
Thank you for your contribution @davidsenk!

PR #194, opened 2024-02-18
Upgrade testng to 7.8.0

gaul:


gaul:
```
class file has wrong version 55.0, should be 52.0
```

gaul:
Requires JDK 11: https://testng.org/#_requirements

PR #196, opened 2024-02-23
JCLOUDS-1627: Upgrade to Jakarta packages

gaul:
This resolves an issue with newer Guice versions.


gaul:
Not using the latest versions since they are not compatible with Java 8.

PR #199, opened 2024-03-18
JCLOUDS-1630: Handle URI template properly with opened curve bracket

Maxim-Gadalov:
https://issues.apache.org/jira/projects/JCLOUDS/issues/JCLOUDS-1630


nacx (approved):
LGTM

PR #200, opened 2024-03-18
JCLOUDS-1631: fix AWSRequestAuthorizeSignatureV4 when query contains special chars

Maxim-Gadalov:
https://issues.apache.org/jira/projects/JCLOUDS/issues/JCLOUDS-1631

### DIFF
--- a/providers/aws-s3/src/test/java/org/jclouds/aws/s3/filter/AwsRequestAuthorizeSignatureV4Test.java
+++ b/providers/aws-s3/src/test/java/org/jclouds/aws/s3/filter/AwsRequestAuthorizeSignatureV4Test.java
@@ -1,3 +1,19 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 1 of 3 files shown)


providers/aws-s3/src/test/java/org/jclouds/aws/s3/filter/AwsRequestAuthorizeSignatureV4Test.java
```diff
@@ -0,0 +1,53 @@
+package org.jclouds.aws.s3.filter;
```
gaul, line 1: Please add license header to resolve CI failure.

gaul (changes requested):

providers/aws-s3/src/test/java/org/jclouds/aws/s3/filter/AwsRequestAuthorizeSignatureV4Test.java
Maxim-Gadalov: added

gaul:
Thank you for your contribution @Maxim-Gadalov!

PR #203, opened 2024-04-03
more validation for containerName and blobKey to avoid access escape

jixinchi:


apis/filesystem/src/main/java/org/jclouds/filesystem/predicates/validators/internal/FilesystemBlobKeyValidatorImpl.java
```diff
@@ -38,6 +38,8 @@ public void validate(String name) throws IllegalArgumentException {
         //blobkey cannot start with / (or \ in Windows) character
         if (name.startsWith("\\") || name.startsWith("/"))
             throw new IllegalArgumentException("Blob key '" + name + "' cannot start with \\ or /");
+        if (name.contains("../"))
+            throw new IllegalArgumentException("Blob key '" + name + "' cannot contains ../");
```
nacx, line 42: super nit
```suggestion
            throw new IllegalArgumentException("Blob key '" + name + "' cannot contain ../");
```

jixinchi: done

apis/filesystem/src/main/java/org/jclouds/filesystem/predicates/validators/internal/FilesystemBlobKeyValidatorImpl.java
```diff
@@ -38,6 +38,8 @@ public void validate(String name) throws IllegalArgumentException {
         //blobkey cannot start with / (or \ in Windows) character
         if (name.startsWith("\\") || name.startsWith("/"))
             throw new IllegalArgumentException("Blob key '" + name + "' cannot start with \\ or /");
+        if (name.contains("../"))
+            throw new IllegalArgumentException("Blob key '" + name + "' cannot contain ../");
```
gaul, line 42: Why is this different than the container name validator?  For robustness should this tokenize the path via `/` then check each component to see if one contains `.` or `..`?  This would allow keys like `..foo` to work.

apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
```diff
@@ -691,6 +698,8 @@ public BlobAccess getBlobAccess(String containerName, String blobName) {
 
    @Override
    public void setBlobAccess(String container, String name, BlobAccess access) {
+      filesystemContainerNameValidator.validate(container);
+      filesystemBlobKeyValidator.validate(name);
```
gaul, line 702: Should this also apply to `getBlobAccess`, `putBlob`, and `removeBlob`?  `getBlobKeysInsideContainer` too I guess.

gaul (changes requested):
Please add some unit test that exercises this logic.

apis/filesystem/src/main/java/org/jclouds/filesystem/strategy/internal/FilesystemStorageStrategyImpl.java
jixinchi: These methods have been applied by the validators. I just added some missing ones.

apis/filesystem/src/main/java/org/jclouds/filesystem/predicates/validators/internal/FilesystemBlobKeyValidatorImpl.java
jixinchi: I think `..foo` should be allowed. I tried several object storage services, such as gcloud and aliyun oss, they all allow `..foo` as object key.

jixinchi:
@gaul Could you please review the new commits?

gaul:
Thank you for your contribution @jixinchi!

PR #204, opened 2024-04-14
JCLOUDS-1635: Add COOL and COLD to Tier

gaul:
The former replaces INFREQUENT.  References gaul/s3proxy#625.


gaul:
@timuralp what do you think?

nacx (approved):

PR #206, opened 2024-07-03
Modify S3 endpoint regex pattern to support VPC endpoints.

jcalcote:
Virtual private clouds are a relatively new feature of AWS. The jclouds AWS endpoint regex was not originally designed to be able to handle VPC formatted endpoints. This new expanded regex does the trick. The change also (incidentally) allows the regex to recognize both forms of basic s3 endpoints: s3-region and s3.region.

### DIFF
--- a/apis/s3/src/main/java/org/jclouds/s3/filters/AwsHostNameUtils.java
+++ b/apis/s3/src/main/java/org/jclouds/s3/filters/AwsHostNameUtils.java
@@ -26,7 +26,7 @@ import java.util.regex.Pattern;
 
 public class AwsHostNameUtils {
 
-   private static final Pattern S3_ENDPOINT_PATTERN = Pattern.compile("^(?:.+\\.)?s3[.-]([a-z0-9-]+)$");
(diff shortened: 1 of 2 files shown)


apis/s3/src/main/java/org/jclouds/s3/filters/AwsHostNameUtils.java
```diff
@@ -26,7 +26,7 @@
 
 public class AwsHostNameUtils {
 
-   private static final Pattern S3_ENDPOINT_PATTERN = Pattern.compile("^(?:.+\\.)?s3[.-]([a-z0-9-]+)$");
+   private static final Pattern S3_ENDPOINT_PATTERN = Pattern.compile("^(?:.+\\.)?s3[.\\-]([a-z0-9-]+)(?>\\.[a-z0-9-]+)*$");
```
gaul, line 29: Can you add some kind of unit test which exercises this?

gaul (changes requested):

apis/s3/src/main/java/org/jclouds/s3/filters/AwsHostNameUtils.java
jcalcote: Done.

gaul:
Thank you for your contribution @jcalcote!

PR #207, opened 2024-07-30
Fix ContextBuilder bug when using credentials supplier with azureblob

maxjaspersmiley:
Azure cloud providers specifically use an identity as part of their endpoint. 

When attempting to use a ContextBuilder with the (new-ish?) credentialsSupplier option, we were seeing errors building an endpoint due to string interpolation failures, since the Azure module was not expecting a Supplier<Credentials> (rather, a specific `identity` and `credential`)

This is not a comprehensive fix (any subsequent change to the identity will not be reflected in the endpoint going forward) but partially addresses a serious problem (unable to use ContextBuilder's `credentialsSupplier` for Azure providers)

I plan on a more thorough change that will allow on-the-fly changes to an Azure provider's identity/credential, but this is more difficult. I believe I'll need to find every place where the endpoint is referenced and handle the case where a credentialsSupplier is being used (IOW, every place where the endpoint must be generated on the fly). 

Any pointers or feedback about the above is appreciated!


gaul:
Thank you for your contribution @maxjaspersmiley!

PR #208, opened 2024-08-02
Fixes issue with AzureBlobStore blockid incorrectly using base64url encoding

twick00:
Despite [JCLOUDS-1615](https://issues.apache.org/jira/browse/JCLOUDS-1615), using base64url encoded values is not supported by the [Put Block](https://learn.microsoft.com/en-us/rest/api/storageservices/put-block) endpoint.

When attempting to query this endpoint with `partNumber` set to `248` (which is `blockid: AAAA-A==`) I received this error:

```xml
<?xml version="1.0" encoding="utf-8"?><Error><Code>InvalidQueryParameterValue</Code><Message>Value for one of the query parameters specified in the request URI is invalid.
RequestId:632a8422-501e-006c-3db4-e238fb000000
Time:2024-07-30T19:10:34.4638223Z</Message><QueryParameterName>blockid</QueryParameterName><QueryParameterValue>AAAA-A==</QueryParameterValue><Reason>Not a valid base64 string.</Reason></Error>
```

I have confirmed directly with Microsoft that the correct value here is a standard base64 string that is url encoded, which should not be confused with base64url encoding which is NOT supported.

As far as the original ticket ([JCLOUDS-1615](https://issues.apache.org/jira/browse/JCLOUDS-1615)) which mentions an error related to  `+` reporting as not a valid base64 string, I have tried to replicate the issue and I am unable to do so. My guess is that this was a bug that existed on Microsoft's side and they have since fixed it. Either way, from their own words, it should be a base64 string. `+` and `/` are supported.

### DIFF
--- a/providers/azureblob/src/main/java/org/jclouds/azureblob/blobstore/AzureBlobStore.java
+++ b/providers/azureblob/src/main/java/org/jclouds/azureblob/blobstore/AzureBlobStore.java
@@ -440,9 +440,10 @@ public class AzureBlobStore extends BaseBlobStore {
       return sync.putBlockList(mpu.containerName(), azureBlob, blocks.build());
    }
 
-   static String makeBlockId(int partNumber) {
-       return BaseEncoding.base64Url().encode(Ints.toByteArray(partNumber));
-   }
+  static String makeBlockId(int partNumber) {
+      // Azure expects a base64-encoded string ONLY. It does not support base64url encoding. 
+     return BaseEncoding.base64().encode(Ints.toByteArray(partNumber));
+  }
    
    @Override
    public MultipartPart uploadMultipartPart(MultipartUpload mpu, int partNumber, Payload payload) {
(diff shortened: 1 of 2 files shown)


providers/azureblob/src/main/java/org/jclouds/azureblob/blobstore/AzureBlobStore.java
```diff
@@ -440,9 +440,10 @@ public String completeMultipartUpload(MultipartUpload mpu, List<MultipartPart> p
       return sync.putBlockList(mpu.containerName(), azureBlob, blocks.build());
    }
 
-   static String makeBlockId(int partNumber) {
-       return BaseEncoding.base64Url().encode(Ints.toByteArray(partNumber));
-   }
+  static String makeBlockId(int partNumber) {
+      // Azure expects a base64-encoded string ONLY. It does not support base64url encoding. 
+     return BaseEncoding.base64().encode(Ints.toByteArray(partNumber));
```
twick00, line 445: The url encoding is handled downstream by `org.jclouds.util.Strings2`'s `urlEncode` method.

twick00:
I know this is not the correct venue for this but I also tried to sign up to report the issue via jclouds' Jira but nobody has responded to that request in the last week. I also tried contacting the IRC mentioned on the [community page](https://jclouds.apache.org/community/) and there is practically nobody in there.

I would love to help contribute more to this project but this seems to be the only possible option right now.

gaul:
Thank you for your contribution @twick00!  I approved the JIRA request.  I only use the mailing list not IRC or Slack.

PR #216, opened 2025-02-12
Fix digitalocean2 NumberFormatException

felfert:
Changed org.jclouds.digitalocean2.domain.Action.id from int to long
 in order to fit large numbers sent in responses.

 This appears to be a misinterpretation of DigitalOceans ApiDocs,
 because in these docs, they fail to specify the size of their
 "integer" type. Therefore, I looked into their official GO Api
 sources at https://github.com/digitalocean/godo. There, they use
 the int type, which is platform dependant and usually has 64 bits.

 Therefore our digitalocean2 provider, should be refactored, changing
 **_all_** Ids from int/Integer to long/Long.

This PR is just the immediate fixup to make the jclouds (jenkins plugin)
work with current DigitalOcean again. The refactoring should be done
ASAP, otherwise we could run into similar problems, when the id values in
their API-resonses grow over time.


felfert:
BTW: This change applies to older versions as well without the need for any
adaptions (tested against 2.5.0 here).

nacx (approved):
LGTM. Thank you!

PR #218, opened 2025-04-10
Fix NPE on failure to parse AWSError when determing whether to retry request

maxjaspersmiley:
Under some conditions, attempting to parse an AWSError from a request (org.jclouds.aws.util.AWSUtils#parseAWSErrorFromContent) can wind up with a null result. 

S3RedirectionRetryHandler was not handling such null responses gracefully, and was instead hitting a NullPointerException. 

This one-line fix changes S3RedirectionRetryHandler to null-check results from the method mentioned above.


nacx (approved):

gaul:
Thank you for your contribution @maxjaspersmiley!

